Issue #8, opened 2019-01-22
Set up API response tests


kcover:
**Describe the task**
Create tests to validate the functionality of the api so far. That is, create tests that confirm the api gives the proper responses when given specific input. The aim of this ticket is not to create all of the necessary tests, but to create enough tests to cover the current functionality of the api so that we can do TDD going forward.

PR #9, opened 2019-01-22
Rep-8 Create API Response Tests

kcover:
#### What does this PR do?
Creates tests to confirm that the API sends the proper responses. It also fixes an issue I found while creating the tests where the hostname and port of an object were returned as null.

@paouelle @peterhuffer @clockard 

Start an instance of DDF and run the tests.

#8 

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
+++ b/distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
@@ -1,0 +1,199 @@
+/**
+ * Copyright (c) Codice Foundation
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ditto.replication.admin.query;
+
+import static com.jayway.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.hasItem;
+import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.core.Is.is;
+
+import com.jayway.restassured.specification.RequestSpecification;
+import java.net.MalformedURLException;
+import org.junit.Test;
+
+public class ITReplicationQuery {
+
+  private static final String GRAPHQL_ENDPOINT = "https://localhost:8993/admin/hub/graphql";
+
+  private static final String URL = "https://localhost:9999";
+
+  // ----------------------------------- Site Tests -----------------------------------//
+
+  @Test
+  public void createReplicationSite() throws MalformedURLException {
+    String name = "create";
+    asAdmin()
+        .body(makeCreateSiteQuery(name, URL))
+        .when()
+        .post(GRAPHQL_ENDPOINT)
+        .then()
+        .statusCode(200)
+        .header("Content-Type", is("application/json;charset=utf-8"))
+        .body("data.createReplicationSite.name", is(name))
+        .body("data.createReplicationSite.address.url", is(URL))
+        .body("data.createReplicationSite.address.host.hostname", is("localhost"))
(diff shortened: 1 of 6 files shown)


distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
```diff
@@ -0,0 +1,190 @@
...
+  public void createReplicationSiteWithIntegerName() {
+    asAdmin()
+        .body(
+            String.format(
+                "{\"query\":\"mutation{ createReplicationSite(name: 25, address: { url: \\\"%s\\\"}){ id name address{ host{ hostname port } url }}}\"}",
+                URL))
+        .when()
+        .post(GRAPHQL_ENDPOINT)
+        .then()
+        .statusCode(200)
```
paouelle: ❗️ I would say returning 200 OK when a request is invalid is a problem. This is not something a client of an API would expect.

distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
```diff
@@ -0,0 +1,190 @@
...
+
+  @Test
+  public void createReplicationSite() throws MalformedURLException {
+    String name = "create";
+    asAdmin()
+        .body(makeCreateSiteQuery(name, URL))
+        .when()
+        .post(GRAPHQL_ENDPOINT)
+        .then()
+        .statusCode(200)
```
paouelle: ❓ Should we verify the content-type of the responses?

distributions/test/pom.xml
```diff
@@ -0,0 +1,31 @@
...
+ * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
+ * version 3 of the License, or any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ *
+ **/
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
paouelle: ❓ Should we follow the DDF convention and define the project hierarchy for tests as: distributions/test/itests?

distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
peterhuffer: When type validation fails, GraphQL returns a 200 but the "errors" field in the JSON response is populated.

distributions/test/pom.xml
kcover: I tried to address this one already. Is there something that I missed?

distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
kcover: I would think that a 400 would make more sense, but I think that is a problem with the endpoint, not with the tests.

distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
kcover: Is it really possible to return anything other than json? And, would that be a test of it's own, or an assertion we make in each test?

paouelle: Is there a content-type header in the response? There should be right?

distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
paouelle: it depends. I agree that 400 makes more sense to me as well. If that is what we say our API should return then we should validate that in our test and let the test fail until we fix it. If it turns out that 200 is valid in this case as we specify an error code inside the response message than the test case is ok. It all depends on what we document our API/protocol to be.

distributions/test/pom.xml
paouelle: Should there be another level underneath `test` named `itests` with its own pom file and the test class defined there?

kcover: There is. The module is declared just below this comment 🙂

paouelle: Oops. I guess I just carried over this comment from the old repo 😆

PR #21, opened 2019-02-10
REP-20 Add scheduling to replication and condense status messages

### LINKED ISSUE #20: Internalize replication scheduling
**Describe the task**
Instead of relying on scheduling command to run replication an internal mechanism should be added to automatically run the replication jobs. This could potentially result in many more replication status entries than before so a way to condense the status objects should also be included.

clockard:
#### What does this PR do?
Add scheduling to replication and condense status messages
Install replication and create a replication configuration verify after 10min or so that there are some status entries via the replication:status command. 

Fixes #20

### DIFF
--- a/replication-api-impl/pom.xml
+++ b/replication-api-impl/pom.xml
@@ -235,6 +235,12 @@
             <artifactId>failsafe</artifactId>
             <version>1.1.0</version>
         </dependency>
+        <dependency>
+            <groupId>ddf.catalog.core</groupId>
+            <artifactId>filter-proxy</artifactId>
+            <version>${ddf.version}</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <build>
         <resources>
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
@@ -20,10 +20,22 @@ import ddf.catalog.data.impl.MetacardImpl;
 import ddf.catalog.data.types.Core;
 import ddf.catalog.filter.FilterBuilder;
 import ddf.catalog.operation.impl.CreateRequestImpl;
+import ddf.catalog.operation.impl.DeleteRequestImpl;
+import ddf.catalog.operation.impl.UpdateRequestImpl;
+import ddf.catalog.source.CatalogProvider;
 import ddf.catalog.source.IngestException;
 import ddf.catalog.source.SourceUnavailableException;
+import ddf.security.service.SecurityServiceException;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import net.jodah.failsafe.Failsafe;
+import net.jodah.failsafe.RetryPolicy;
+import org.codice.ddf.security.common.Security;
+import org.codice.ditto.replication.api.ReplicationPersistenceException;
 import org.codice.ditto.replication.api.ReplicationStatus;
 import org.codice.ditto.replication.api.ReplicatorHistory;
 import org.codice.ditto.replication.api.Status;
@@ -32,11 +44,16 @@ import org.codice.ditto.replication.api.mcard.ReplicationHistory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+/** A metacard/catalog based persistence implementation of the ReplicatorHistory interface */
 public class ReplicatorHistoryImpl implements ReplicatorHistory {
 
   private static final Logger LOGGER = LoggerFactory.getLogger(ReplicatorHistoryImpl.class);
 
+  private final Security security;
+
   private final CatalogFramework framework;
+
+  private final CatalogProvider provider;
 
   private final FilterBuilder filterBuilder;
 
@@ -46,13 +63,41 @@ public class ReplicatorHistoryImpl implements ReplicatorHistory {
 
   public ReplicatorHistoryImpl(
       CatalogFramework framework,
+      CatalogProvider provider,
       FilterBuilder filterBuilder,
       MetacardHelper helper,
       MetacardType metacardType) {
+    this(framework, provider, filterBuilder, helper, metacardType, Security.getInstance());
+  }
+
+  /*Visible for testing only*/
+  ReplicatorHistoryImpl(
+      CatalogFramework framework,
+      CatalogProvider provider,
+      FilterBuilder filterBuilder,
+      MetacardHelper helper,
+      MetacardType metacardType,
+      Security security) {
     this.framework = framework;
+    this.provider = provider;
     this.filterBuilder = filterBuilder;
     this.helper = helper;
     this.metacardType = metacardType;
+    this.security = security;
+  }
+
+  /** Initialize the history records. This is used to update outdated formats */
+  public void init() {
+    // Using failsafe retry here because when this is deployed
+    // as a kar file there were several times when the catalog
+    // was not available so re-running was required
+    RetryPolicy retryPolicy =
+        new RetryPolicy()
+            .withDelay(5, TimeUnit.SECONDS)
+            .withMaxDuration(2, TimeUnit.MINUTES)
+            .retryOn(Exception.class);
+
+    Failsafe.with(retryPolicy).run(this::runResultsAsSystemUser);
   }
 
   @Override
@@ -82,12 +127,46 @@ public class ReplicatorHistoryImpl implements ReplicatorHistory {
   @Override
   public void addReplicationEvent(ReplicationStatus replicationStatus) {
     try {
+      ReplicationStatus previous =
+          getReplicationEvents(replicationStatus.getReplicatorName())
+              .stream()
+              .findFirst()
+              .orElse(null);
+      if (previous != null) {
+        addStats(previous, replicationStatus);
+        provider.update(new UpdateRequestImpl(previous.getId(), getMetacardFromStatus(previous)));
+        return;
+      }
+      replicationStatus.setLastRun(replicationStatus.getStartTime());
+      if (replicationStatus.getStatus().equals(Status.SUCCESS)) {
+        replicationStatus.setLastSuccess(replicationStatus.getStartTime());
+      }
       framework.create(new CreateRequestImpl(getMetacardFromStatus(replicationStatus)));
+
     } catch (IngestException | SourceUnavailableException e) {
-      LOGGER.warn(
-          "Error creating replication history item for {}",
-          replicationStatus.getReplicatorName(),
+      throw new ReplicationPersistenceException(
+          "Error creating replication history item for " + replicationStatus.getReplicatorName(),
           e);
+    }
+  }
+
+  @Override
+  public void removeReplicationEvent(ReplicationStatus replicationStatus) {
+    try {
+      provider.delete(new DeleteRequestImpl(replicationStatus.getId()));
+    } catch (IngestException e) {
+      throw new ReplicationPersistenceException(
+          "Error deleting replication history item " + replicationStatus.getReplicatorName(), e);
+    }
+  }
+
+  @Override
+  public void removeReplicationEvents(Set<String> ids) {
+    try {
+      provider.delete(new DeleteRequestImpl(ids.toArray(new String[ids.size()])));
+    } catch (IngestException e) {
+      throw new ReplicationPersistenceException(
+          "Error deleting replication history items " + ids, e);
     }
   }
 
@@ -116,6 +195,10 @@ public class ReplicatorHistoryImpl implements ReplicatorHistory {
                 metacard, ReplicationHistory.STATUS, Status.PENDING.name())));
     status.setStartTime(
         helper.getAttributeValueOrDefault(metacard, ReplicationHistory.START_TIME, null));
+    status.setLastSuccess(
+        helper.getAttributeValueOrDefault(metacard, ReplicationHistory.LAST_SUCCESS, null));
+    status.setLastRun(
+        helper.getAttributeValueOrDefault(metacard, ReplicationHistory.LAST_RUN, null));
 
     return status;
   }
@@ -134,8 +217,91 @@ public class ReplicatorHistoryImpl implements ReplicatorHistory {
     helper.setIfPresent(mcard, ReplicationHistory.PULL_BYTES, replicationStatus.getPullBytes());
     helper.setIfPresent(mcard, ReplicationHistory.PUSH_BYTES, replicationStatus.getPushBytes());
     helper.setIfPresent(mcard, ReplicationHistory.STATUS, replicationStatus.getStatus().name());
+    helper.setIfPresent(mcard, ReplicationHistory.LAST_SUCCESS, replicationStatus.getLastSuccess());
+    helper.setIfPresent(mcard, ReplicationHistory.LAST_RUN, replicationStatus.getLastRun());
     mcard.setId(replicationStatus.getId());
     mcard.setTags(Collections.singleton(ReplicationHistory.METACARD_TAG));
     return mcard;
+  }
+
+  private void addStats(ReplicationStatus base, ReplicationStatus status) {
+    base.setPushCount(base.getPushCount() + status.getPushCount());
+    base.setPushBytes(base.getPushBytes() + status.getPushBytes());
+    base.setPushFailCount(base.getPushFailCount() + status.getPushFailCount());
+    base.setPullCount(base.getPullCount() + status.getPullCount());
+    base.setPullBytes(base.getPullBytes() + status.getPullBytes());
+    base.setPullFailCount(base.getPullFailCount() + status.getPullFailCount());
+    if (base.getLastRun() == null || status.getStartTime().after(base.getLastRun())) {
+      base.setLastRun(status.getStartTime());
+    }
+    base.setStatus(status.getStatus());
+    if (base.getStartTime().after(status.getStartTime())) {
+      base.setStartTime(status.getStartTime());
+    }
+    if (status.getPushCount() > 0 || status.getPullCount() > 0) {
+      base.setDuration(base.getDuration() + status.getDuration());
+    }
+
+    if (status.getStatus().equals(Status.SUCCESS)
+        && (base.getLastSuccess() == null || status.getStartTime().after(base.getLastSuccess()))) {
+      base.setLastSuccess(status.getStartTime());
+    }
+  }
+
+  /** */
+  void runResultsAsSystemUser() {
+    security.runAsAdmin(
+        () -> {
+          try {
+            security.runWithSubjectOrElevate(
+                () -> {
+                  getReplicationEvents()
+                      .stream()
+                      .map(ReplicationStatus::getReplicatorName)
+                      .distinct()
+                      .map(this::getReplicationEvents)
+                      .forEach(this::condenseResults);
+                  return null;
+                });
+          } catch (SecurityServiceException | InvocationTargetException e) {
+            LOGGER.debug("Error condensing replication history.", e);
+          }
+          return null;
+        });
+  }
+
+  /**
+   * Condenses old mvp status events into a single event
+   *
+   * @param events events to condense for a single configuration events should be in order of newest
+   *     to oldest
+   */
+  void condenseResults(List<ReplicationStatus> events) {
+    if (events.size() == 1 && events.get(0).getLastRun() != null) {
+      return;
+    }
+    ReplicationStatus condensedStatus = null;
+    Set<String> idsToRemove = new HashSet<>();
+    for (ReplicationStatus status : events) {
+      idsToRemove.add(status.getId());
+      if (condensedStatus == null) {
+        condensedStatus = new ReplicationStatus(status.getReplicatorName());
+        condensedStatus.setStartTime(status.getStartTime());
+        condensedStatus.setStatus(status.getStatus());
+      }
+      addStats(condensedStatus, status);
+    }
+
+    if (condensedStatus == null) {
+      return;
+    }
+
+    try {
+      framework.create(new CreateRequestImpl(getMetacardFromStatus(condensedStatus)));
+      removeReplicationEvents(idsToRemove);
+    } catch (IngestException | SourceUnavailableException e) {
+      LOGGER.debug(
+          "Error creating replication history item for {}", condensedStatus.getReplicatorName(), e);
+    }
   }
 }
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorImpl.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorImpl.java
@@ -106,6 +106,7 @@ public class ReplicatorImpl implements Replicator {
             } catch (InterruptedException e) {
               LOGGER.trace("InterruptedException in executor. This is expected during shutdown.");
               Thread.currentThread().interrupt();
+              break;
             }
           }
         });
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
@@ -1,0 +1,113 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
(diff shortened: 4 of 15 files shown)


replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
```diff
@@ -208,7 +208,11 @@ private Filter buildFilter() {
     }
 
     if (lastSuccessfulRun != null) {
-      Date timeStamp = new Date(lastSuccessfulRun.getStartTime().getTime() - 1000);
+      long time = lastSuccessfulRun.getStartTime().getTime();
+      if (lastSuccessfulRun.getLastSuccess() != null) {
+        time = lastSuccessfulRun.getLastSuccess().getTime();
+      }
+      Date timeStamp = new Date(time - 1000);
```
paouelle: ❓ Why the `- 1000`? Comment maybe?

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicationStatus.java
```diff
@@ -24,10 +24,14 @@
 
   private Date startTime;
 
+  private Date lastSuccess;
```
paouelle: ✏️ Annotate with `@Nullable`

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicationStatus.java
```diff
@@ -66,6 +70,14 @@ public void setStartTime(Date startTime) {
     this.startTime = startTime;
   }
 
+  public Date getLastSuccess() {
```
paouelle: ✏️ Annotate with `@Nullable`

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicationStatus.java
```diff
@@ -66,6 +70,14 @@ public void setStartTime(Date startTime) {
...
+    return lastSuccess;
+  }
+
+  public void setLastSuccess(Date lastSuccess) {
```
paouelle: ✏️ Annotate `lastSuccess` with `@Nullable`

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicatorHistory.java
```diff
@@ -14,6 +14,7 @@
 package org.codice.ditto.replication.api;
 
 import java.util.List;
+import java.util.Set;
```
paouelle: ✏️ class documentation

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicatorHistory.java
```diff
@@ -38,4 +39,8 @@
    * @param replicationStatus ReplicationConfig event to store
    */
   void addReplicationEvent(ReplicationStatus replicationStatus);
+
+  void removeReplicationEvent(String replicatorid);
```
paouelle: ✏️ method documentation

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicatorHistory.java
```diff
@@ -38,4 +39,8 @@
...
+
+  void removeReplicationEvents(Set<String> ids);
```
paouelle: ✏️ method documentation

replication-api/src/main/java/org/codice/ditto/replication/api/StatusPeriod.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ditto.replication.api;
+
+public enum StatusPeriod {
```
paouelle: ✏️ class documentation

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/StatusCondenser.java
```diff
@@ -0,0 +1,183 @@
...
+            replicatorHistory
+                .getReplicationEvents(config.getName())
+                .stream()
+                .filter(s -> s.getPeriod().equals(StatusPeriod.WEEK))
+                .collect(Collectors.toList()));
+    condenseResults(events, StatusPeriod.YEAR, 365);
+  }
+
+  /**
+   * Condenses events into new events of the given type based on the given window
```
kcover: Could you be more specific? This one was pretty hard for me to grasp.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/StatusCondenser.java
```diff
@@ -0,0 +1,183 @@
...
+        oldest.setStartTime(startOfDay(status.getStartTime()));
+        cutoff = new Date(oldest.getStartTime().getTime() + TimeUnit.DAYS.toMillis(daysInWindow));
+      }
+
+      if (status.getStartTime().before(cutoff)) {
+        addStats(oldest, status);
+        idsToRemove.add(status.getId());
+      } else {
+        // save oldest day as a week
+        replicatorHistory.addReplicationEvent(oldest);
```
kcover: ❓ Couldn't this also be the oldest week saved as a year? Above you seem to pass in different time units.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/StatusCondenser.java
```diff
@@ -0,0 +1,183 @@
...
+
+  /**
+   * Condenses events into new events of the given type based on the given window
+   *
+   * @param events events to condense. Events must be ordered oldest to newest
+   * @param outType the output event type
+   * @param daysInWindow number of days in the outputType
+   */
+  void condenseResults(List<ReplicationStatus> events, StatusPeriod outType, int daysInWindow) {
+    ReplicationStatus oldest = null;
```
kcover: ✏️ Since this is going to be continuously modified, should we give it another name to better explain what it will hold later? Maybe call it base, condensed, or result.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/StatusCondenser.java
```diff
@@ -0,0 +1,183 @@
...
+                .collect(Collectors.toList()));
+    condenseResults(events, StatusPeriod.WEEK, 7);
+    events =
+        Lists.reverse(
+            replicatorHistory
+                .getReplicationEvents(config.getName())
+                .stream()
+                .filter(s -> s.getPeriod().equals(StatusPeriod.WEEK))
+                .collect(Collectors.toList()));
+    condenseResults(events, StatusPeriod.YEAR, 365);
```
paouelle: ❗️ what about leap years?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/StatusCondenser.java
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  /**
+   * Condenses events into new events of the given type based on the given window
+   *
+   * @param events events to condense. Events must be ordered oldest to newest
+   * @param outType the output event type
+   * @param daysInWindow number of days in the outputType
+   */
+  void condenseResults(List<ReplicationStatus> events, StatusPeriod outType, int daysInWindow) {
```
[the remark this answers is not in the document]
paouelle: ❗️ It also looks to me like the boundaries are actually based on when the method is called instead of being fixed in time. Users would expect that all stats for the first week of the year would be collected together, and so on and not the next 7 days from when it is called. The same for year, ...

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
```diff
@@ -0,0 +1,103 @@
...
+    this.security = security;
+  }
+
+  public void init() {
+    long period =
+        Long.parseLong(
+            System.getProperty(
+                "org.codice.replication.period", String.valueOf(TimeUnit.MINUTES.toSeconds(5))));
+    scheduledExecutor.scheduleAtFixedRate(
+        this::scheduleReplicationAsSystemUser,
```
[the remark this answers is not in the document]
clockard: Done

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/StatusCondenser.java
paouelle: We should talk with @mcalcote about how we aggregate as it might become confusing to the user especially when it comes down to display aggregated values to users.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/StatusCondenser.java
```diff
@@ -0,0 +1,183 @@
...
+    ReplicationStatus oldest = null;
+    Date cutoff = null;
+    Set<String> idsToRemove = new HashSet<>();
+    for (ReplicationStatus status : events) {
+      if (oldest == null) {
+        oldest = new ReplicationStatus(status.getReplicatorName());
+        oldest.setPeriod(outType);
+        oldest.setStartTime(startOfDay(status.getStartTime()));
+        cutoff = new Date(oldest.getStartTime().getTime() + TimeUnit.DAYS.toMillis(daysInWindow));
+      }
```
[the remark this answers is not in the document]
paouelle: 🤔

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
paouelle: this is why commenting code is sometimes important 😉

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
```diff
@@ -91,6 +99,24 @@ public void addReplicationEvent(ReplicationStatus replicationStatus) {
     }
   }
 
+  @Override
+  public void removeReplicationEvent(String replicatorid) {
+    try {
+      provider.delete(new DeleteRequestImpl(replicatorid));
+    } catch (IngestException e) {
+      LOGGER.debug("Error deleting replication history item {}", replicatorid, e);
```
paouelle: FYI, you could have left the log at error and simply added another log at debug with the exception in order to get the stack trace. My comment was just to not log the exception at anything but debug.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/StatusCondenser.java
```diff
@@ -0,0 +1,183 @@
...
+  private void addStats(ReplicationStatus base, ReplicationStatus status) {
+    base.setPushCount(base.getPushCount() + status.getPushCount());
+    base.setPushBytes(base.getPushBytes() + status.getPushBytes());
+    base.setPushFailCount(base.getPushFailCount() + status.getPushFailCount());
+    base.setPullCount(base.getPullCount() + status.getPullCount());
+    base.setPullBytes(base.getPullBytes() + status.getPullBytes());
+    base.setPushFailCount(base.getPushFailCount() + status.getPushFailCount());
+    base.setDuration(base.getDuration() + status.getDuration());
+    base.setStatus(
+        base.getStatus().equals(Status.SUCCESS) || status.getStatus().equals(Status.SUCCESS)
```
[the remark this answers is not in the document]
paouelle: Should we create a ticket in the backlog to track that?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
```diff
@@ -134,8 +215,86 @@ private Metacard getMetacardFromStatus(ReplicationStatus replicationStatus) {
...
+    if (base.getLastRun() == null || status.getStartTime().after(base.getLastRun())) {
+      base.setLastRun(status.getStartTime());
+    }
+    base.setStatus(status.getStatus());
+    if (base.getStartTime().after(status.getStartTime())) {
+      base.setStartTime(status.getStartTime());
+    }
+    if (status.getStatus().equals(Status.SUCCESS)) {
+      if (status.getPushCount() > 0 || status.getPullCount() > 0) {
+        base.setDuration(base.getDuration() + status.getDuration());
```
peterhuffer: ❓From the API, we will expose this as "total run time" for the configuration, but would something like "last run duration" be more useful, ie don't accumulate these? I'm thinking the more runs, the less meaningful "total run time" becomes.

replication-api-impl/pom.xml
```diff
@@ -235,6 +235,12 @@
             <artifactId>failsafe</artifactId>
```
peterhuffer: ✏️ We should update the coverage numbers in this pom.

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicatorHistory.java
```diff
@@ -14,7 +14,9 @@
 package org.codice.ditto.replication.api;
 
 import java.util.List;
+import java.util.Set;
 
+/** Persistence interface for storing, relieving, and deleting history items */
```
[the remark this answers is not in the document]
clockard: Fixed

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicationPersistenceException.java
```diff
@@ -0,0 +1,15 @@
+package org.codice.ditto.replication.api;
```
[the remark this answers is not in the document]
clockard: Good catch! Looks like we don't have a check in our build for that at the moment.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
clockard: I was thinking this would be useful still for calculating the average transfer rate but you have a good point that it might be also good to know what the last run duration was. Maybe also last bytes/item transferred. Would you be ok with exploring that in a later change?

peterhuffer: Yeah I'm okay with changing it later. I'm not sure if average transfer rate would be useful either though. I'd think they would want to know the transfer rate of the currently running job.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
```diff
@@ -36,8 +48,12 @@
```
paouelle: ✏️ Class documentation and copyright notice

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
paouelle: ❓ Shouldn't we add the duration time even if it failed? In a failure, wouldn't there still have been records replicated potentially?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
```diff
@@ -134,8 +215,90 @@ private Metacard getMetacardFromStatus(ReplicationStatus replicationStatus) {
...
+    security.runAsAdmin(
+        () -> {
+          try {
+            security.runWithSubjectOrElevate(
+                () -> {
+                  getReplicationEvents()
+                      .stream()
+                      .map(ReplicationStatus::getReplicatorName)
+                      .distinct()
+                      .forEach(config -> condenseResults(getReplicationEvents(config)));
```
paouelle:
```suggestion
                     .map(this::getReplicationEvents)
                     .forEach(this::condenseResults);
```

paouelle: ❓ Are we not creating a new record each time we are condensing instead of creating one final one at the end of condensing all those we are iterating here?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorImpl.java
```diff
@@ -106,6 +106,7 @@ public void init() {
             } catch (InterruptedException e) {
```
paouelle: ✏️ The above block nested inside the `execute()` should be moved to a private method

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
```diff
@@ -0,0 +1,115 @@
...
+    this.replicator = notNull(replicator);
+    this.replicatorConfigLoader = notNull(replicatorConfigLoader);
+    this.security = security;
+  }
+
+  public void init() {
+    long period =
+        Long.parseLong(
+            System.getProperty(
+                "org.codice.replication.period", String.valueOf(TimeUnit.MINUTES.toSeconds(5))));
```
paouelle: ❓ Should we define `5` as a constant? Actually the whole string representation of 5 minutes in seconds should probably be defined as a constant.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
```diff
@@ -0,0 +1,115 @@
...
+    scheduledExecutor.scheduleAtFixedRate(
+        this::replicateAsSystemUser, TimeUnit.MINUTES.toSeconds(1), period, TimeUnit.SECONDS);
```
paouelle: ✏️ 1 minute should probably be defined as a constant for clarity

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
```diff
@@ -0,0 +1,115 @@
...
+        .forEach(this::submitSyncRequest);
+  }
+
+  private void submitSyncRequest(SyncRequest request) {
+    try {
+      LOGGER.debug("Submitting replication request for {} ", request.getConfig().getName());
+      replicator.submitSyncRequest(request);
+    } catch (InterruptedException e) {
+      LOGGER.debug("Exception received from the replicator. ", e);
+      Thread.currentThread().interrupt();
```
paouelle: ❗️ The problem in doing this here is that you are in the middle of an iteration controlled by `scheduleReplication()` which means that the next iteration will be interrupted again. Typically, at a low level like this it is better to throw back an `InterruptedException` and let the code where the loop is handle it properly. Since in your case you want to abort the loop (I assume). The calling code would catch the exception outside of the loop and propagate back the interruption state on the thread before bailing out.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
paouelle: ❓ Won't these records (saved at the same time) be picked up by the next poll?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/mcard/ReplicationHistoryAttributes.java
```diff
@@ -38,6 +38,22 @@
             false /* tokenized */,
             false /* multivalued */,
             BasicTypes.DATE_TYPE));
+    descriptors.add(
```
paouelle: ✏️ Class documentation and copyright notice. Maybe double check all files for that.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
```diff
@@ -46,13 +62,40 @@
...
   public ReplicatorHistoryImpl(
       CatalogFramework framework,
+      CatalogProvider provider,
       FilterBuilder filterBuilder,
       MetacardHelper helper,
       MetacardType metacardType) {
+    this(framework, provider, filterBuilder, helper, metacardType, Security.getInstance());
+  }
+
+  public ReplicatorHistoryImpl(
```
mcalcote: Only for testing?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
```diff
@@ -82,15 +125,49 @@ public ReplicatorHistoryImpl(
...
-          "Error creating replication history item for {}",
-          replicationStatus.getReplicatorName(),
+      throw new ReplicationPersistenceException(
+          "Error creating replication history item for " + replicationStatus.getReplicatorName(),
           e);
     }
   }
 
+  @Override
+  public void removeReplicationEvent(String replicatorid) {
```
mcalcote: ✏️ Maybe this is really nit picky, but I would prefer if this took a ReplicationStatus argument

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
```diff
@@ -82,15 +125,49 @@ public ReplicatorHistoryImpl(
...
+    try {
+      provider.delete(new DeleteRequestImpl(replicatorid));
+    } catch (IngestException e) {
+      throw new ReplicationPersistenceException(
+          "Error deleting replication history item " + replicatorid, e);
```
mcalcote: If we had the ReplicationStatus, this message could utilize the replicatorName instead of a UUID and be more useful

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
```diff
@@ -82,15 +125,49 @@ public ReplicatorHistoryImpl(
...
+    }
+  }
+
+  @Override
+  public void removeReplicationEvents(Set<String> ids) {
+    try {
```
mcalcote: if you keep the current implementation, you could pass each id to removeReplicationEvent and reduce this method

mcalcote: Ignore 👆 if you change removeReplicationEvent

clockard: This delete will perform much better than calling the other remove method. I think I am going to just change the other method.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
clockard: Updated

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
clockard: Updated

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
clockard: Added a comment that this is only for testing. Made it package private as well.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/mcard/ReplicationHistoryAttributes.java
clockard: Looks like this class does have the copyright header. Will add class documentation.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
clockard: Reworked calling method to fix this

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
clockard: Added constant

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
clockard: Added constant

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
clockard: Added class documentation

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryImpl.java
clockard: Yup. Updated

PR #23, opened 2019-02-13
REP-22 Implement Create Operation for RepSyncs

### LINKED ISSUE #22: Implement Create Operation For RepSyncs
**Describe the task**
RepSyncs are a configuration that define what, when, and where data will be replicated to and from. Currently the create operation is mocked out in the API. The mock should be replaced with a functioning method that will persist RepSyncs to the back-end for later retrieval.

kcover:
#### What does this PR do?
This PR refactors the replicationPersistentStore.

#### Who is reviewing it?
@clockard @peterhuffer @paouelle 

#### How should this be tested?
Test creating, running, deleting, and suspending replication configs in the UI and CLI.

Fixes #22 

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests


replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/AbstractPersistable.java
```diff
@@ -0,0 +1,88 @@
...
+    if (id == null) {
+      setId(UUID.randomUUID().toString());
+    } else {
+      setId(id);
+    }
+  }
+
+  protected AbstractPersistable(String type, Map<String, Object> properties) {
+    this.type = type;
+    setProperties(properties);
```
paouelle: ❗️ We shouldn't be calling an abstract method from the constructor. This implies that you would be calling into subclass' code before the subclass has been initialized.

replication-api/pom.xml
```diff
@@ -52,6 +57,61 @@
...
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <!--
+                                    When overriding the limits in child pom files make sure
```
peterhuffer: ✏️ We should put the .75 limits in the top level pom.
❓ Are we able to use the LenientLimit class?

replication-api/src/main/java/org/codice/ditto/replication/api/modern/ReplicationPersistentStore.java
```diff
@@ -0,0 +1,55 @@
...
+   * @return A list of Objects that were retrieved from storage
+   */
+  <T extends Persistable> List<T> get(Class<T> klass, String typeName);
+
+  /**
+   * Saves the persistable object
+   *
+   * @param object The object to save
+   */
+  void save(Persistable object);
```
peterhuffer: ❓ What if it fails to be saved?

replication-api/src/test/java/org/codice/ditto/replication/api/PersistableTest.java
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testCreatePersistableThrowsException() {
+    // Trying to invoke a constructor on the Persistable interface will cause an Exception
+    Persistable.createPersistable(Persistable.class, new HashMap<>());
+  }
+
+  // public so reflection can use constructor
+  public static class testPersistable implements Persistable {
```
peterhuffer: ✏️ `TestPersistable`

replication-api/src/main/java/org/codice/ditto/replication/api/modern/ReplicationPersistentStore.java
```diff
@@ -0,0 +1,55 @@
...
+   * creates an instance of class {@code <T>} with the data retrieved from persistence.
+   *
+   * @param klass The class of object to instantiate and return.
+   * @param typeName The persistence type of the desired object.
+   * @param id The id of the object to search for.
+   * @param <T> A class that implements the Persistable interface. An object of this type will be
+   *     created from the persisted data and returned.
+   * @return The object that was retrieved from storage
+   */
+  <T extends Persistable> Optional<T> get(Class<T> klass, String typeName, String id);
```
peterhuffer: ❓ What if we are unable to retrieve the objects from storage?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/AbstractPersistable.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+public abstract class AbstractPersistable implements Persistable {
+
+  private static final String ID = "id";
+
+  private String type;
+
+  protected Map<String, Object> properties;
+
+  protected AbstractPersistable(@Nullable String id, String type) {
```
peterhuffer: ✏️ Might be more clear to have an additional separate ctor that only takes `type` and generate an ID there rather than forcing the caller to pass in null to generate an ID.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/AbstractPersistable.java
```diff
@@ -0,0 +1,88 @@
...
+    return new HashMap<>(properties);
+  }
+
+  /**
+   * This method should be implemented by subclasses to confirm that the given properties have all
+   * the required fields for that class before setting the properties as their own.
+   *
+   * @param properties the properties the AbstractPersistable should hold
+   */
+  public abstract void setProperties(Map<String, Object> properties);
```
peterhuffer: ❓ What if a property is the wrong expected type?

replication-api/src/main/java/org/codice/ditto/replication/api/modern/Persistable.java
```diff
@@ -0,0 +1,71 @@
...
+   * constructor that takes in a {@code Map<String, Object>} as its sole argument. Otherwise a call
+   * to this method with that class as an argument will throw an {@link IllegalArgumentException}.
+   *
+   * @param klass The class of object to create an instance of.
+   * @param properties The properties that the created object will have.
+   * @param <T> A class implementing the persistable interface, whose constructor will be called and
+   *     the created object returned.
+   * @return An instance of T that has been created with the given properties.
+   */
+  static <T extends Persistable> T createPersistable(
```
peterhuffer: This seems weird because the caller is required to know an implementation of Persistable. I can't just fetch a `ReplicationSite` right? I have to fetch a `ReplicationSiteImpl`. It seems like it should be `ReplicationSite implements Persistable`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/RepSyncImpl.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  public void setFilter(String filter) {
+    properties.put(FILTER, filter);
+  }
+
+  @Override
+  public void setProperties(Map<String, Object> properties) {
+    if (!hasRequiredFields(properties)) {
+      throw new IllegalArgumentException("Null values are not allowed.");
```
paouelle: ✏️ If there is an expectation that an implementation throws an `IllegalArgumetException` for a certain reason than this should be documented in the parent class along with the contract for this method

paouelle: ❓ What about if the values for specific keys are of the wrong type? Deserialization, or here validation, of the object should ensure the object is completely valid before allowing the object to be used.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Map;
+import java.util.Optional;
+import org.codice.ddf.persistence.PersistenceException;
+import org.codice.ddf.persistence.PersistentStore;
+import org.codice.ditto.replication.api.modern.Persistable;
+import org.codice.ditto.replication.api.modern.ReplicationPersistentStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ReplicationPersistentStoreImpl implements ReplicationPersistentStore {
```
paouelle: ✏️ Document the class

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/RepSyncImpl.java
```diff
@@ -0,0 +1,98 @@
...
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ditto.replication.api.impl.modern;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.codice.ditto.replication.api.modern.RepSync;
+
+public class RepSyncImpl extends AbstractPersistable implements RepSync {
```
paouelle: ✏️ Document the class

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+  private static final Logger LOGGER =
+      LoggerFactory.getLogger(ReplicationPersistentStoreImpl.class);
+
+  private static final int DEFAULT_PAGE_SIZE = 100;
+
+  private static final int DEFAULT_START_INDEX = 0;
+
+  private PersistentStore persistentStore;
```
paouelle:
```suggestion
  private final PersistentStore persistentStore;
```

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,96 @@
...
+
+  public ReplicationPersistentStoreImpl(PersistentStore persistentStore) {
+    this.persistentStore = persistentStore;
+  }
+
+  @Override
+  public <T extends Persistable> Optional<T> get(Class<T> klass, String typeName, String id) {
+    String cql = String.format("'id' = '%s'", id);
```
paouelle:
```suggestion
    final String cql = String.format("'id' = '%s'", id);
```

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    List<Map<String, Object>> storedMaps = Collections.emptyList();
+    try {
+      storedMaps = persistentStore.get(typeName, cql, DEFAULT_START_INDEX, DEFAULT_PAGE_SIZE);
+    } catch (PersistenceException e) {
+      LOGGER.debug(
+          "An error occurred while retrieving a {} with an ID of {}", klass.getName(), id, e);
```
paouelle: ✏️ This would result in leaving the map empty and the next debug trace to come out before returning `Optional.empty()`. Should we just return here? That being said, I am still wondering why we are not propagating the error wrapped into a new one specific to our API to the client's of this method to let them know you are unable to retrieve anything as opposed to telling them there is nothing for that id.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+  }
+
+  @Override
+  public <T extends Persistable> List<T> get(Class<T> klass, String typeName) {
+    List<Map<String, Object>> storedMaps = Collections.emptyList();
+    try {
+      storedMaps = persistentStore.get(typeName, "", DEFAULT_START_INDEX, DEFAULT_PAGE_SIZE);
+    } catch (PersistenceException e) {
+      LOGGER.debug("An error occurred while retrieving data of type {}.", typeName, e);
```
paouelle: ❓ Why not returned directly with an `Collections.emptyList()` or an empty list?

The whole method could be simplified as:

```
try {
    try {
      return persistentStore.get(typeName, "", DEFAULT_START_INDEX, DEFAULT_PAGE_SIZE)
        .stream()
        .map(m -> Persistable.createPersistable(kclass, m))
        .collect(Collectors.toList());
    } catch (PersistenceException e) {
      LOGGER.debug("An error occurred while retrieving data of type {}.", typeName, e);
      return new ArrayList<>(); // or Collections.emptyList();
   }
```

Same comment as above, exceptions should not result in an empty result. Empty means there was nothing in the persistence store whereas as an error is an indication that you cannot tell. The client should be handling that on its own.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,96 @@
...
+      LOGGER.debug(
+          "An error occurred while retrieving a {} with an ID of {}", klass.getName(), id, e);
+    }
+
+    if (storedMaps.isEmpty()) {
+      LOGGER.debug("Couldn't find persisted object of type {} with ID {}", typeName, id);
+      return Optional.empty();
+    } else {
+      if (storedMaps.size() > 1) {
+        LOGGER.debug("Found two objects in persistence of type {} with ID {}", typeName, id);
```
paouelle: ❓ Is that even possible?

The whole method could be simplified to:

```
try {
  return persistentStore.get(typeName, cql, DEFAULT_START_INDEX, DEFAULT_PAGE_SIZE)
    .stream()
    .limit(1)
    .map(m -> Persistable.createPersistable(klass, m))
    .findFirst();
} catch (PersistenceException e) {
      LOGGER.debug(
          "An error occurred while retrieving a {} with an ID of {}", klass.getName(), id, e);
      return Option.empty();
}
```

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    return results;
+  }
+
+  @Override
+  public void save(Persistable object) {
+    String persistenceType = object.getType();
+    try {
+      persistentStore.add(persistenceType, object.getProperties());
+    } catch (PersistenceException e) {
+      LOGGER.debug("An error occurred while saving data of type {}.", persistenceType);
```
paouelle: ❗️ You are letting the client think it worked even though it failed

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,96 @@
...
+    for (Map<String, Object> result : storedMaps) {
+      results.add(Persistable.createPersistable(klass, result));
+    }
+
+    return results;
+  }
+
+  @Override
+  public void save(Persistable object) {
+    String persistenceType = object.getType();
```
paouelle:
```suggestion
    final String persistenceType = object.getType();
```

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/AbstractPersistable.java
```diff
@@ -0,0 +1,88 @@
...
+
+  public String getType() {
+    return type;
+  }
+
+  public void setType(String type) {
+    this.type = type;
+  }
+
+  public Map<String, Object> getProperties() {
```
paouelle: ❗️ This should probably be package private since it is only required by the persistence manager and therefore could return the map as is as opposed to cloning it. Worst case scenario, could you not just return an unmodifiable version of it?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationSiteImpl.java
```diff
@@ -13,47 +13,64 @@
  */
 package org.codice.ditto.replication.api.impl.modern;
 
-import java.net.URL;
+import java.util.HashMap;
+import java.util.Map;
 import org.codice.ditto.replication.api.modern.ReplicationSite;
 
-public class ReplicationSiteImpl implements ReplicationSite {
+public class ReplicationSiteImpl extends AbstractPersistable implements ReplicationSite {
```
paouelle: ✏️ document the class

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationSiteImpl.java
```diff
@@ -13,47 +13,64 @@
...
   }
 
   public void setId(String id) {
-    this.id = id;
+    super.setId(id);
+  }
+
+  @Override
+  public void setProperties(Map<String, Object> properties) {
+    if (properties.get(NAME) == null || properties.get(URL) == null) {
```
paouelle: ☝️ See previous comments about validating the types of each fields too.

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/AbstractPersistableTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+  private static String TEST_ID = "testId";
+
+  private static String TYPE = "test";
+
+  private TestPersistable persistable;
+
+  @Before
+  public void setup() {
+    persistable = new TestPersistable(TEST_ID, TYPE);
```
paouelle: ❓ Why not just initialize `persistable` above directly?

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/AbstractPersistableTest.java
```diff
@@ -0,0 +1,102 @@
...
+  @Test
+  public void testGetters() {
+    Map<String, Object> map = new HashMap<>();
+    map.put("id", TEST_ID);
+    assertThat(persistable.getProperties(), equalTo(map));
+    assertThat(persistable.getId(), equalTo(TEST_ID));
+  }
+
+  @Test
+  public void testSetters() {
```
paouelle: ✏️ This is 2 test cases

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/RepSyncImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+  private static final String TEST_SOURCE = "source";
+
+  private static final String TEST_DESTINATION = "destination";
+
+  private static final String TEST_FILTER = "filter";
+
+  private RepSyncImpl repSync;
+
+  @Before
+  public void setup() {
```
paouelle: ❓ Why the need for a setup method instead of just initializing `repSync` above?

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/RepSyncImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+    Map<String, Object> map = new HashMap<>();
+    map.put("id", TEST_ID);
+    map.put("name", TEST_NAME);
+    map.put("source", TEST_SOURCE);
+    map.put("destination", TEST_DESTINATION);
+    map.put("filter", TEST_FILTER);
+
+    repSync = new RepSyncImpl(map);
+
+    assertThat(repSync.getId(), equalTo(TEST_ID));
```
paouelle: technically speaking, you should also be verifying all other properties

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/RepSyncImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+  public void testPropertyConstructorMissingName() {
+    Map<String, Object> map = new HashMap<>();
+    map.put("id", TEST_ID);
+    map.put("source", TEST_SOURCE);
+    map.put("destination", TEST_DESTINATION);
+    map.put("filter", TEST_FILTER);
+
+    repSync = new RepSyncImpl(map);
+
+    assertThat(repSync.getId(), equalTo(TEST_ID));
```
paouelle: ❓ Are you expecting to get here?

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/RepSyncImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testPropertyConstructorMissingValues() {
+    new RepSyncImpl(new HashMap<>());
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
paouelle: ❓ How do you know this exception is because of the missing name?

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/RepSyncImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+    map.put("destination", TEST_DESTINATION);
+    map.put("filter", TEST_FILTER);
+
+    repSync = new RepSyncImpl(map);
+
+    assertThat(repSync.getId(), equalTo(TEST_ID));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testPropertyConstructorMissingSource() {
```
paouelle: ☝️ Same comments as above

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/RepSyncImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+    map.put("destination", TEST_DESTINATION);
+    map.put("filter", TEST_FILTER);
+
+    repSync = new RepSyncImpl(map);
+
+    assertThat(repSync.getId(), equalTo(TEST_ID));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testPropertyConstructorMissingDestination() {
```
paouelle: ☝️ Same comments as above

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/RepSyncImplTest.java
```diff
@@ -0,0 +1,145 @@
...
+    map.put("source", TEST_SOURCE);
+    map.put("filter", TEST_FILTER);
+
+    repSync = new RepSyncImpl(map);
+
+    assertThat(repSync.getId(), equalTo(TEST_ID));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testPropertyConstructorMissingFilter() {
```
paouelle: ☝️ Same comments as above

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/ReplicationSiteImplTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+    assertThat(site.getId(), equalTo(TEST_ID));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void testPropertyConstructorWithMissingValues() {
+    new ReplicationSiteImpl(new HashMap<>());
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
paouelle: ❓ How do you know it is because of a missing name?

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/ReplicationSiteImplTest.java
```diff
@@ -0,0 +1,93 @@
...
+  public void testPropertyConstructorWithMissingName() {
+    Map<String, Object> map = new HashMap<>();
+    map.put("id", TEST_ID);
+    map.put("url", TEST_URL);
+    site = new ReplicationSiteImpl(map);
+  }
+
+  @Test(expected = IllegalArgumentException.class)
```
paouelle: ❓ How do you know it is because of a missing url?

replication-api/src/main/java/org/codice/ditto/replication/api/modern/Persistable.java
paouelle: ❗️ Agreed. Such a method should not be in the API at all. It should be defined in a service and allow the implementation to register the service. In addition, identification should be done via the object's interface and not the implementation's class.

replication-api/src/main/java/org/codice/ditto/replication/api/modern/ReplicationPersistentStore.java
paouelle: Agreed. As commented before, such distinction is required from a client's perspective

replication-api/src/main/java/org/codice/ditto/replication/api/modern/ReplicationPersistentStore.java
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Attempts to find an object that has been saved with the given persistence type and ID. It then
+   * creates an instance of class {@code <T>} with the data retrieved from persistence.
+   *
+   * @param klass The class of object to instantiate and return.
+   * @param typeName The persistence type of the desired object.
+   * @param id The id of the object to search for.
+   * @param <T> A class that implements the Persistable interface. An object of this type will be
+   *     created from the persisted data and returned.
+   * @return The object that was retrieved from storage
```
paouelle: ✏️ "... or empty if the object does not exist"

replication-api/src/main/java/org/codice/ditto/replication/api/modern/ReplicationPersistentStore.java
```diff
@@ -0,0 +1,55 @@
...
+   * Attempts to retrieve all objects saved with the given persistence type. It then creates a list
+   * of objects of class {@code <T>} with the data retrieved from persistence.
+   *
+   * @param klass The class of object to instantiate and return.
+   * @param typeName The persistence type of the desired object.
+   * @param <T> A class that implements the Persistable interface. Objects of this type will be
+   *     created from the persisted data and returned.
+   * @return A list of Objects that were retrieved from storage
+   */
+  <T extends Persistable> List<T> get(Class<T> klass, String typeName);
```
paouelle: ❓ How about returning a stream instead and let the client decide if they actually need a collection or not? This avoids creating temporary collections for no reasons.

replication-api/src/main/java/org/codice/ditto/replication/api/modern/ReplicationPersistentStore.java
paouelle: Shouldn't the save operation be directly on the object instead? It seems more natural to tell the object to save itself from an OO perspective

replication-api/src/main/java/org/codice/ditto/replication/api/modern/ReplicationPersistentStore.java
```diff
@@ -0,0 +1,55 @@
...
+
+  /**
+   * Saves the persistable object
+   *
+   * @param object The object to save
+   */
+  void save(Persistable object);
+
+  boolean delete(String id);
```
paouelle: ☝️ Same comment

replication-api/src/main/java/org/codice/ditto/replication/api/modern/ReplicationPersistentStore.java
paouelle: ❓ Why both a class and a type? Are those not linked by design? If a client specifies a class, don't we know for fact which type he/she wants? Plus it now requires the client to know all the types. Where do we define these? Inside the class?

replication-api/src/main/java/org/codice/ditto/replication/api/modern/ReplicationPersistentStore.java
paouelle: ☝️ Same comment as before with respect to type and class

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/AbstractPersistable.java
kcover: I'm not really seeing what might go wrong but I can set the properties in the constructors of the subclasses

replication-api/pom.xml
kcover: .75 limits are already in the top level pom and yea I'll change this to use LenientLimit

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,160 @@
...
+    try {
+      return (String) FieldUtils.readStaticField(klass, "PERSISTENCE_TYPE");
+    } catch (Exception e) {
+      throw new ReplicationPersistenceException(
+          "Failed to retrieve the persistence type of the class " + klass.getName(), e);
+    }
+  }
+
+  @Override
+  public void save(Persistable object) throws ReplicationPersistenceException {
```
kcover: An Idea that makes a lot of sense to me is that we should do all of our validation of data as it comes into our system. So I have validation in the _get_ methods. Should I be performing validation here as well? Obviously I don't want to save invalid data, but as long as we have proper validation elsewhere we shouldn't have to worry about it right?

paouelle: There are typically 2 levels of validation you see:
1- per attribute validation (e.g. timeout must be positive)
2- per object validation (e.g. if a given attribute is true then another one must be provided with something); in other words attribute dependencies between each other which from the valid state of an object.

I like the idea of validation as the data comes into the system. The question would be can we avoid doing that each time someone calls a getter?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/modern/RepSyncImpl.java
```diff
@@ -0,0 +1,128 @@
...
+import com.google.common.annotations.VisibleForTesting;
+import java.util.HashMap;
+import java.util.Map;
+import org.codice.ditto.replication.api.modern.RepSync;
+
+/**
+ * RepSyncImpl represents a RepSync and has methods that allow it to easily be converted to, or
+ * from, a map.
+ */
+public class RepSyncImpl extends AbstractPersistable implements RepSync {
```
peterhuffer: ❓ This is going to be a little difficult from the GraphQL API since now we have the configuration separated between `RepSync` and `ReplicatorConfig`. I'm wondering if we should just combine them now?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/RepSyncImpl.java
```diff
@@ -0,0 +1,185 @@
...
+import org.codice.ditto.replication.api.ReplicatorConfig;
+
+/**
+ * RepSyncImpl is an implementation of ReplicatorConfig and has methods that allow it to easily be
+ * converted to, or from, a map.
+ */
+public class RepSyncImpl extends AbstractPersistable implements ReplicatorConfig {
+
+  // public so that the persistent store can access it using reflection
+  public static String PERSISTENCE_TYPE = "replication_repsync";
```
kcover: I think I might just hard code getType() to return this string and the others like it. That way, in the ReplicationPersistentStore, we can just create an instance of this class and get the type from it without using reflection.

paouelle: Keep in mind that you would be creating an object for no other reasons than to get its type. Using reflection is not an evil thing to do; it is a valid and acceptable practice in many APIs today.

kcover:
build now

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/RepSyncImpl.java
kcover: My only concern was that when we use reflection our tools can't help us, Intellij can't remind me that I need to add a static variable with this name if I create another persistable, and I'd like to leverage our tools as much as possible. I'm going to leave this alone however because it's not a big deal.

replication-api/src/main/java/org/codice/ditto/replication/api/data/ReplicationSite.java
```diff
@@ -11,31 +11,36 @@
...
 
   /**
-   * Get the human readable name of this site
+   * Set the name of this site
    *
-   * @return site name
+   * @param name the name to give this site
    */
-  String getName();
+  void setName(String name);
```
peterhuffer: ✏️  Since SiteLoader has a method `saveSite(String name, String url)`, these setters and the `createPersistable(Class<T> klass)` seem to duplicate the same behavior.

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicatorConfig.java
```diff
@@ -46,27 +56,78 @@
...
+   * @param destinationId the id for the destination site
+   */
+  void setDestination(String destinationId);
+
   /**
    * Get the cql filter used for determining the data set to replicate
    *
    * @return replication cql filter
    */
+  @Deprecated
```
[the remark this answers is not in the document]
kcover: This was a case of me trying to avoid making changes in a lot of places. I wanted to refactor it to be called getFilter but didn't want to mess with all the files it was being used in. So, I just added a getFilter method and deprecated the old one. Looking back on all the other refactoring I did, I think I'll just go back and rename it.

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicationPersistentStore.java
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * Instantiates an implementation of the given interface which the client will then be responsible
+   * for populating with the proper fields.
+   *
+   * @param klass The interface to instantiate an implementation for
+   * @param <T> An interface that extends persistable
+   * @return An instance of an implementation of the specified interface
+   */
+  <T extends Persistable> T createPersistable(Class<T> klass);
```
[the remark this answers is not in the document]
kcover: This basically comes back to whether we want setters in our interfaces right? I get the desire for small interfaces but having the setters on there makes it easy for us to update an object. If I want to edit a sites name I can just load it up, set the name, and save it, instead of creating a whole new site and saving that. If we do some refactoring this will also let us avoid having a dependency on the api and the impl componenents, we can just depend on the api.

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicationPersistentStore.java
```diff
@@ -0,0 +1,72 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ditto.replication.api.persistence;
+
+import java.util.Optional;
+import java.util.stream.Stream;
+import org.codice.ditto.replication.api.data.Persistable;
+
+/** A ReplicationPersistentStore performs CRUD operations for replication. */
+public interface ReplicationPersistentStore {
```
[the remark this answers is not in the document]
kcover: I think they should probably all document that exception. Nice catch.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -60,12 +74,37 @@ public ReplicatorConfigImpl(ReplicatorConfig config) {
...
-    this.id = id;
+  @Override
+  public void readFromMap(Map<String, Object> properties) {
+    setId((String) properties.get(ID_KEY));
+    setName((String) properties.get(NAME_KEY));
+    setSource((String) properties.get(SOURCE_KEY));
+    setDestination((String) properties.get(DESTINATION_KEY));
+    setFilter((String) properties.get(FILTER_KEY));
+    Object biDirectionalObject = properties.get(BIDIRECTIONAL_KEY);
+    if (biDirectionalObject instanceof String) {
```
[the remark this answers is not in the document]
kcover: The persistentItem class we use in the PersistentStore doesn't hold booleans so it converts it to a string. It then gets handed back as a string and I can't think of a good way to keep that implementation detail in the ReplicationPersistentStore.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -60,12 +74,37 @@ public ReplicatorConfigImpl(ReplicatorConfig config) {
...
+
+  @Override
+  public Map<String, Object> writeToMap() {
+    Map<String, Object> result = new HashMap<>();
+    result.put(ID_KEY, getId());
+    result.put(NAME_KEY, getName());
+    result.put(SOURCE_KEY, getSource());
+    result.put(DESTINATION_KEY, getDestination());
+    result.put(FILTER_KEY, getFilter());
+    result.put(BIDIRECTIONAL_KEY, isBiDirectional());
```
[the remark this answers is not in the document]
kcover: I agree but I wanted to avoid refactoring everything that uses direction instead of bidirectional for now.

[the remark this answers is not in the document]
kcover: I think I'll just add a TODO here for now.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -109,6 +148,26 @@ public void setDestination(String destination) {
...
+  public boolean isBiDirectional() {
+    return direction == Direction.BOTH;
+  }
+
+  @Override
+  public void setBiDirectional(boolean biDirectional) {
+    direction = biDirectional ? Direction.BOTH : Direction.PUSH;
+  }
+
   public void setReplicationType(ReplicationType type) {
```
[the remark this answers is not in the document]
kcover: Another refactor I was putting off. I'll add a TODO

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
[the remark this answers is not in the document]
paouelle: If that is the case than do bother putting a boolean i nthere and make it consistent between the serialization and deserialization.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,170 @@
...
+      storedMaps = persistentStore.get(typeName, cql, DEFAULT_START_INDEX, DEFAULT_PAGE_SIZE);
+    } catch (PersistenceException e) {
+      throw new ReplicationPersistenceException(
+          String.format(
+              "An error occurred while trying to retrieve an object from persistence with class %s and id %s",
+              klass.getName(), id),
+          e);
+    }
+
+    if (storedMaps.isEmpty()) {
```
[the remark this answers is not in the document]
kcover: This is more refactoring that I think should be done outside of this ticket. Maybe I should just make a ticket that points to this PR for things to refactor?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,170 @@
...
+      LOGGER.debug("Couldn't find persisted object of type {} with ID {}", typeName, id);
+      return Optional.empty();
+    } else {
+      if (storedMaps.size() > 1) {
+        LOGGER.debug("Found two objects in persistence of type {} with ID {}", typeName, id);
```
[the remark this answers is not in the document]
kcover: This should never happen. Solr doesn't allow two objects to have the same ID and I would hope that whatever else we might use for storage wouldn't allow it either. Should I just remove this? Or do you think I should throw a runtime exception?

replication-api/pom.xml
```diff
@@ -52,6 +57,61 @@
...
+                                    to override all three limits. Limits that are excluded
+                                    will be set to 0 not 0.75
+                                    -->
+                                    <limits>
+                                        <limit>
```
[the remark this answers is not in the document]
kcover: Do we even need the jacoco plugin here? It's the api, it doesn't have any tests. I'm not sure how it's even building if we have a .04 minimum coverage.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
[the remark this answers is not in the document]
kcover: the PersistentItem class is in ddf and I don't think we want to change it if we're moving away from ddf.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStoreImpl.java
```diff
@@ -0,0 +1,170 @@
...
+  @Override
+  public void save(Persistable object) {
+    AbstractPersistable persistable = (AbstractPersistable) object;
+    String persistenceType = persistable.getType();
+    PersistentItem persistentItem = getPersistentItemFromPersistable(persistable);
+
+    try {
+      persistentStore.add(persistenceType, persistentItem);
+    } catch (PersistenceException e) {
+      throw new ReplicationPersistenceException(
```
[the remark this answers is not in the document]
kcover: oops I think I meant to add the persistenceException as the cause.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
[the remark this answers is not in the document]
paouelle: I am not suggesting modifying the PersistentItem. I am suggesting making sure your serialization is consistent with your deserialization. If you know booleans don't work ... than don't put any booleans to start with. Store something else (e.g. a string, an enum string, ...) in the PersistentItem such that you know you will be getting it back properly

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -46,7 +60,7 @@
   public ReplicatorConfigImpl() {}
 
   public ReplicatorConfigImpl(ReplicatorConfig config) {
-    this.id = config.getId();
+    this.setId(config.getId());
```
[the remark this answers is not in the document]
kcover: I don't think we need this copy constructor anymore if we are going to have setters on our interfaces. I'm going to remove it.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
```diff
@@ -0,0 +1,77 @@
...
+
+  @Override
+  public String getType() {
+    return PERSISTENCE_TYPE;
+  }
+
+  @Override
+  public Map<String, Object> writeToMap() {
+    Map<String, Object> result = new HashMap<>();
+    result.put(ID_KEY, getId());
```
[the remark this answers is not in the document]
kcover: How about I just create the map in the base class so it doesn't need to take one in?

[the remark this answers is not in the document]
paouelle: That would work too. You would then be adding to the version retrieved from `super`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
```diff
@@ -0,0 +1,77 @@
...
+  private static final String NAME_KEY = "name";
+
+  private static final String URL_KEY = "url";
+
+  private String name;
+
+  private String url;
+
+  @Override
+  public String getName() {
```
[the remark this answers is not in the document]
kcover: Name will only be common to configs and sites and I don't really want to create another abstract class just to handle the name variable for two subclasses. But I can if we think it's necessary

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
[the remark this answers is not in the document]
kcover: I created an issue that will encompass this change https://github.com/connexta/replication/issues/36

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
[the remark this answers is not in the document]
kcover: This should also be taken care of in #36

kcover:
One more commit coming up to refactor the return type of the _get by id_ methods in the persistent store and the managers to move away from using optionals.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -132,37 +138,43 @@ public ReplicationField updateReplication(
                             + name
                             + ". No configuration found."));
 
-    ReplicatorConfigImpl config = new ReplicatorConfigImpl(existingConfig);
-    config.setName(name);
-    config.setSource(sourceId);
-    config.setDestination(destinationId);
-    config.setCql(filter);
-    config.setDirection(biDirectional ? Direction.BOTH : Direction.PUSH);
```
clockard: ❗️ Right now ReplicatorImpl still only uses the direction field. We either need to set it here or update the ReplicatorImpl to also look at the bi-directional field.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -0,0 +1,81 @@
...
+
+  public abstract String getType();
+
+  /**
+   * Writes the variables of the persistable to a map. Any implementation of this method in a
+   * subclass should first make a call to the super version before performing its own functionality.
+   *
+   * @return a map containing the variable of the persistable
+   */
+  public Map<String, Object> writeToMap() {
```
clockard: Should writeToMap and readFromMap be methods defined in Persistable? Maybe also think about renaming the methods to something like `toMap` and `fromMap`. The read and write part make it not as clear in my mind.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -13,15 +13,33 @@
...
+
+  private static final String DESTINATION_KEY = "destination";
+
+  private static final String FILTER_KEY = "filter";
+
+  private static final String RETRY_COUNT_KEY = "retry_count";
+
+  private static final String BIDIRECTIONAL_KEY = "bidirectional";
+
+  // TODO: do we want to keep descriptions?
```
clockard: I think we do for now. Can we remove this comment?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStore.java
```diff
@@ -0,0 +1,153 @@
...
+
+  private static final int DEFAULT_START_INDEX = 0;
+
+  private final PersistentStore persistentStore;
+
+  public ReplicationPersistentStore(PersistentStore persistentStore) {
+    this.persistentStore = persistentStore;
+  }
+
+  private <T extends AbstractPersistable> T createPersistable(Class<T> klass) {
```
clockard: I think this class could just deal with `T extends Persistable` if you move the to/from map methods to persistable which seems cleaner to me.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorConfigManagerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @Override
+  public List<ReplicatorConfig> getAllConfigs() {
+    return persistentStore.objects(ReplicatorConfigImpl.class).collect(Collectors.toList());
+  }
+
+  @Override
+  public void saveConfig(ReplicatorConfig replicationConfig) {
+    if (replicationConfig instanceof ReplicatorConfigImpl) {
```
clockard: Is this check and cast only needed because of the AbstractPersistable that the impl uses? Would be nice if we didn't have to check and cast here.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/SiteManagerImpl.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  @Override
+  public Set<ReplicationSite> getSites() {
+    return persistentStore.objects(ReplicationSiteImpl.class).collect(Collectors.toSet());
+  }
+
+  @Override
+  public ReplicationSite saveSite(ReplicationSite site) {
+    if (site instanceof ReplicationSiteImpl) {
```
clockard: Same comment about the check and cast

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorConfigManagerImplTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+  @Test
+  public void saveConfig() {
+    loader.saveConfig(new ReplicatorConfigImpl());
+    verify(persistentStore).save(any(ReplicatorConfigImpl.class));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void saveConfigBadConfig() {
+    loader.saveConfig(new BadConfig());
```
clockard: Can you just mock out the config and return one bad value to achieve this instead of implementing the whole thing?

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/persistence/SiteManagerImplTest.java
```diff
@@ -0,0 +1,222 @@
...
+  public void saveSite() {
+    ReplicationSite site = store.saveSite("name", "url");
+    UUID.fromString(site.getId()); // throws exception if an id wasn't generated
+    assertThat(site.getName(), is("name"));
+    assertThat(site.getUrl(), is("url"));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void saveSiteBadSite() {
+    ReplicationSite site = new BadSite();
```
clockard: Same comment as the BadConfig

replication-commands/src/main/java/org/codice/ditto/replication/commands/ConfigEnableCommand.java
```diff
@@ -43,23 +42,22 @@
...
-  @Reference ReplicatorConfigLoader replicatorConfigLoader;
+  @Reference ReplicatorConfigManager replicatorConfigManager;
 
   @Reference Replicator replicator;
 
   @Override
   protected Object executeWithSubject() throws Exception {
     for (String config : configNames) {
-      ReplicatorConfig replicatorConfig = replicatorConfigLoader.getConfig(config).orElse(null);
+      ReplicatorConfig replicatorConfig = replicatorConfigManager.getConfig(config).orElse(null);
```
clockard: These the config here is just the config name. I think you changed the getConfig to only work with the ID so this will need to be changed as well.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
kcover: ReplicatorImpl still works because were not using the bidirectional field just yet. I'm aiming to make that change in #36. Here's what the bidirectional methods currently look like 

` @Override
    this.cql = cql;
  public boolean isBiDirectional() {
    return direction == Direction.BOTH;
  }

  @Override
  public void setBiDirectional(boolean biDirectional) {
    direction = biDirectional ? Direction.BOTH : Direction.PUSH;
  }`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
paouelle: I don't think they should be. We do not want to bleed our implementation to the public API.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStore.java
paouelle: As mentioned above, we do not want to bleed the implementation into the API. Having the to/from on the impl side is appropriate.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorConfigManagerImpl.java
paouelle: The only way to avoid that cast is to move the `save()` directly into the `Persistable`. Then on the implementation side, you can implement without a need to check.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
kcover: I can see the arguments for either way so I think I'm going to stick with the easy route of not changing the persistable interface. I will change the names to the ones you suggested however.

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/data/AbstractPersistableTest.java
```diff
@@ -0,0 +1,55 @@
...
+    AbstractPersistable persistable = new TestPersistable();
+    persistable.setId(id);
+    assertThat(persistable.getId(), equalTo(id));
+  }
+
+  @Test
+  public void generateId() {
+    AbstractPersistable persistable = new TestPersistable();
+    persistable.generateId();
+    UUID.fromString(persistable.getId()); // will throw an exception if a valid id wasn't generated
```
peterhuffer: ❓ This test would fail for `ReplicatorConfig`s since we strip the `-` characters, does that mean we are saving that ID incorrectly? Should we mention this format in the API?

kcover: @clockard do you know why we remove the `-` from our ids? Do we still need to do that?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
```diff
@@ -57,19 +57,24 @@
...
     this.filterBuilder = filterBuilder;
     this.helper = helper;
     this.configMetacardType = configMetacardType;
-    this.siteStore = siteStore;
+    this.siteManager = siteManager;
+  }
+
+  @Override
+  public ReplicatorConfig createConfig() {
+    return new ReplicatorConfigImpl();
```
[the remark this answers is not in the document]
kcover: I think that's a good idea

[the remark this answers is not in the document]
paouelle: I would actually suggest doing it directly in the AbstractPersistable base class default constructor.

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/data/AbstractPersistableTest.java
clockard: We had to do that when saving them as metacards since that is the way the catalog framework does it. I don't think there is a need to do it now.

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/SiteManager.java
```diff
@@ -49,6 +63,7 @@
    * @param name the name of the site to save
    * @param url the url of the site to save
    * @return a boolean indicating whether the save was successful
+   * @throws ReplicationPersistenceException if an error occurs while trying to save the site
    */
   ReplicationSite saveSite(String name, String url);
```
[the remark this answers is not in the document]
kcover: It creates and saves. If another site has the same name then the graphql code should catch that and not attempt to save the site at all.

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/SiteManager.java
```diff
@@ -40,6 +51,9 @@
    *
```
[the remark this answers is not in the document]
kcover: Clarified in the dataManager interface and duplicate names should be detected at the boundary and rejected i.e. the graphql and command line components should handle that

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicatorConfigManager.java
```diff
@@ -39,13 +50,17 @@
    * one. Else, saves the {@link ReplicatorConfig}.
    *
    * @param replicationConfig The configuration to save or update
+   * @throws ReplicationPersistenceException if an error occurs while trying to save the config
+   * @throws IllegalArgumentException if the ReplicatorConfig implementation is not one that can be
+   *     saved
    */
   void saveConfig(ReplicatorConfig replicationConfig);
```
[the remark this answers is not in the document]
kcover: if it has the same name as another config that should be recognized and the save request should be rejected at the boundary. I don't think we want to try and handle that everywhere. If it has the same id as another config then this will overwrite that config.

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/SiteManager.java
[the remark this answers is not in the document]
paouelle: hum! 
How do we handle simultaneous admins updating different configs and naming 2 different the same way? The boundary can only validate that one administrator is not using the same name in 2 places.

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicatorConfigManager.java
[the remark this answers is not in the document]
paouelle: Same reply as before, the boundary can only validate for one admin at a time. If 2 admins are updating configs then boundary checking is no longer valid and the database can be corrupted with duplicate names.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Attempts to populate the variables of the persistable by reading them from the given map. Any
+   * implementation of this method in a subclass should first make a call to the super version
+   * before performing its own functionality.
+   *
+   * @param properties the properties to read into the persistable's variables
+   */
+  public void readFromMap(Map<String, Object> properties) {
+    setId((String) properties.get(ID_KEY));
```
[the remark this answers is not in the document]
kcover: 😬 whoops

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -38,7 +39,8 @@
 
   private static final String SUSPENDED_KEY = "suspended";
 
-  public static final int CURRENT_VERSION = 3;
+  /** 1 - initial version. */
+  public static final int CURRENT_VERSION = 1;
```
kcover: I've changed this back to 1 since it this will be the first version of this config that we save in the persistent store. Any previous version of config can be identified because it will be saved in the ddf catalog.

kcover:
build now

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
clockard: I think this still needs to be 3 since it is the third version of the config. How it is persisted shouldn't really matter to this version.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
```diff
@@ -0,0 +1,78 @@
...
+
+  /** 1 - initial version. */
+  public static final int CURRENT_VERSION = 1;
+
+  private String name;
+
+  private String url;
+
+  public ReplicationSiteImpl() {
+    super();
```
paouelle: ✏️ no need for `super();`; it is the default

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -13,66 +13,92 @@
...
-    this.type = config.getReplicationType();
-    this.source = config.getSource();
-    this.destination = config.getDestination();
-    this.cql = config.getCql();
-    this.description = config.getDescription();
-    this.failureRetryCount = config.getFailureRetryCount();
-    this.suspended = config.isSuspended();
-    this.version = config.getVersion();
+  public ReplicatorConfigImpl() {
+    super();
```
paouelle: ✏️ no need for `super();`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStore.java
```diff
@@ -0,0 +1,156 @@
...
+          "An error occurred while saving data of type " + persistenceType, e);
+    }
+  }
+
+  public void delete(Class<? extends AbstractPersistable> klass, String id) {
+    String cql = String.format("'id' = '%s'", id);
+    try {
+      boolean successful = persistentStore.delete(getPersistenceType(klass), cql) > 0;
+      if (!successful) {
+        throw new NotFoundException(
```
paouelle: ✏️ comment for later related to a previous one. I think we should just return successfully here instead of throwing a not found. At the end of the day, if it ain't found then the end result is the same as what is requested. The operation becomes a no-op as we were asked to delete the object and it is already deleted.

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/DataManager.java
```diff
@@ -0,0 +1,53 @@
...
+   * @throws IllegalArgumentException if the T implementation is not one that can be saved
+   */
+  void save(T object);
+
+  /**
+   * Deletes a T object with the given id
+   *
+   * @param id The id of the object to be removed
+   * @throws ReplicationPersistenceException if an error occurs while trying to delete the config
+   * @throws NotFoundException if an object with the given id cannot be found
```
paouelle: ✏️ Related comment for later ... not sure we should throw back a not found but instead just let it succeed

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/DataManager.java
```diff
@@ -0,0 +1,53 @@
+package org.codice.ditto.replication.api.persistence;
+
+import java.util.stream.Stream;
+import org.codice.ditto.replication.api.data.Persistable;
+
+public interface DataManager<T extends Persistable> {
```
paouelle: ✏️ Class documentation missing

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicatorConfigManager.java
```diff
@@ -0,0 +1,18 @@
...
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ditto.replication.api.persistence;
+
+import org.codice.ditto.replication.api.data.ReplicatorConfig;
+
+public interface ReplicatorConfigManager extends DataManager<ReplicatorConfig> {}
```
paouelle: ✏️ missing javadoc for the class

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
```diff
@@ -144,7 +144,6 @@ public void saveConfig(ReplicatorConfig replicationConfig) {
     }
   }
 
-  @Override
   public void removeConfig(ReplicatorConfig replicationConfig) {
     try {
       framework.delete(new DeleteRequestImpl(replicationConfig.getId()));
```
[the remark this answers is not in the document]
kcover: I'll add the id to the logs in the PR for #36 since I'm doing a lot of work in this class anyway.

peterhuffer:
Successful hero ✔️ .

PR #24, opened 2019-02-16
Add config suspension and active job cancelation

clockard:
#### What does this PR do?

-Adds suspension and version to replication configs
 -- includes new command to modify the the suspension flag of configurations
- Adds ability to cancel currently running sync requests

This is built on top of my other pr so disregard the first REP-20 commit


replication-commands/src/main/java/org/codice/ditto/replication/commands/ConfigEnableCommand.java
```diff
@@ -0,0 +1,76 @@
...
+    for (String config : configNames) {
+      ReplicatorConfig replicatorConfig = replicatorConfigLoader.getConfig(config).orElse(null);
+      if (replicatorConfig == null) {
+        printErrorMessage("No config found for name " + config);
+        continue;
+      }
+      ReplicatorConfigImpl newConfig = new ReplicatorConfigImpl(replicatorConfig);
+      newConfig.setSuspended(suspend);
+      replicatorConfigLoader.saveConfig(newConfig);
+      if (suspend) {
```
peterhuffer: ❓ Could we hide this suspend logic behind an interface? I'm thinking we will have to duplicate this exact logic when this functionality is exposed through the GraphQL endpoint.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorImpl.java
```diff
@@ -245,6 +288,16 @@ public void submitSyncRequest(final SyncRequest syncRequest) throws InterruptedE
     }
   }
 
+  @Override
+  public void cancelSyncRequest(SyncRequest syncRequest) {
+    pendingSyncRequests.remove(syncRequest);
```
peterhuffer: ❓ Will this work for SyncRequest implementations that do not override equals and hashcode?

clockard: I think this is fine for the time being. If we get to the point where we might have multiple implementations then we can revisit this.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorImpl.java
```diff
@@ -186,10 +218,21 @@ private void executeSyncRequest(final SyncRequest syncRequest) {
...
+            syncHelperMap.remove(syncRequest.getConfig().getId());
           }
         });
   }
 
+  SyncHelper getSyncHelper(
+      ReplicationStore source,
+      ReplicationStore destination,
+      ReplicatorConfig config,
+      ReplicationPersistentStore persistentStore,
```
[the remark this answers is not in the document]
clockard: Nope. We can pull from the class variables. Good catch.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorImpl.java
```diff
@@ -186,10 +218,21 @@ private void executeSyncRequest(final SyncRequest syncRequest) {
             status.setStatus(failureStatus);
           } finally {
             completeActiveSyncRequest(syncRequest, status);
+            syncHelperMap.remove(syncRequest.getConfig().getId());
           }
         });
   }
 
+  SyncHelper getSyncHelper(
```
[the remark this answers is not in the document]
clockard: Yup, changing.

replication-commands/src/main/java/org/codice/ditto/replication/commands/ConfigEnableCommand.java
clockard: Good point. I think I will move this into the replicator interface such that there is also a `cancelSynRequest(String configId)`

replication-api/src/main/java/org/codice/ditto/replication/api/Replicator.java
```diff
@@ -25,6 +25,20 @@
    */
   void submitSyncRequest(final SyncRequest syncRequest) throws InterruptedException;
 
+  /**
+   * Cancel a pending or active {@link SyncRequest}
+   *
+   * @param syncRequest
+   */
+  void cancelSyncRequest(final SyncRequest syncRequest);
```
peterhuffer: ❓ Do we need this one now?

replication-api/src/main/java/org/codice/ditto/replication/api/Replicator.java
```diff
@@ -25,6 +25,20 @@
...
+
+  /**
+   * Cancel a pending or active {@link SyncRequest} based on the request configuration id
```
peterhuffer: ✏️ `Cancel all pending or active {@link SyncRequest}...`

admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/persist/SuspendReplication.java
```diff
@@ -0,0 +1,76 @@
...
+  private PidField id;
+
+  private BooleanField suspend;
+
+  private ReplicationUtils replicationUtils;
+
+  public SuspendReplication(ReplicationUtils replicationUtils) {
+    super(FIELD_NAME, DESCRIPTION);
+    this.replicationUtils = replicationUtils;
+    id = new PidField("id");
```
peterhuffer: ✏️ The pid field should be required.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/persist/SuspendReplication.java
```diff
@@ -0,0 +1,76 @@
...
+    suspend = new BooleanField("suspend");
```
peterhuffer: ❓ Should we make this argument required? If we don't, passing in only a pid could suspend the identified config by default, but we should mention it in the description if so.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/persist/CancelReplication.java
```diff
@@ -0,0 +1,73 @@
...
+  public static final BooleanField RETURN_TYPE = new BooleanField();
+
+  private PidField id;
+
+  private ReplicationUtils replicationUtils;
+
+  public CancelReplication(ReplicationUtils replicationUtils) {
+    super(FIELD_NAME, DESCRIPTION);
+    this.replicationUtils = replicationUtils;
+    id = new PidField("id");
```
peterhuffer: ✏️ This pid argument should be required.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/persist/SuspendReplication.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.List;
+import java.util.Set;
+import org.codice.ddf.admin.api.Field;
+import org.codice.ddf.admin.api.fields.FunctionField;
+import org.codice.ddf.admin.common.fields.base.BaseFunctionField;
+import org.codice.ddf.admin.common.fields.base.scalar.BooleanField;
+import org.codice.ddf.admin.common.fields.common.PidField;
+import org.codice.ditto.replication.admin.query.ReplicationUtils;
+
+public class SuspendReplication extends BaseFunctionField<BooleanField> {
```
peterhuffer: ✏️ We should add validation to the suspend and cancel actions for whether or not the id passed in exists, and if not sending back an error.

PR #39, opened 2019-03-15
REP-36 Add functionality to convert old configs

### LINKED ISSUE #36: Add functionality to convert metacard replication configs to the latest version
**Describe the task**
Add functionality to the MetacardConfigLoader to convert any version-less and version 2 configs to the current version and save them in the ReplicatorConfigManager. This should all be done on startup. This task should also remove any remaining code in our data model that reflects the old version of configs. For example, Code mentioning replication types or direction, should refactored or removed.

kcover:
#### What does this PR do?
Turns the MetacardConfigLoader into a bean that will look for old
metacard replication configs saved in the catalog, convert them into the
latest version, and move them into the ReplicationPersistentStore.

#### Who is reviewing it
@peterhuffer @paouelle @clockard @mcalcote 

**edit:** I recommend going through this process with both an 0.2.0 and an earlier version as converting configs from either will hit different branches of the code.

- Start ddf and install replication 0.2.0 or earlier
- create some configs
- uninstall replication
- install the replication SNAPSHOT version
- verify that the old configs appear in the UI with the expected information
- search the catalog for metacards with metacard-tags like "replication-config"
- verify that the old configs are no longer saved in the catalog

Fixes #36

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests


replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
```diff
@@ -54,97 +51,67 @@
...
 
-  public ReplicatorConfig createConfig() {
-    return new ReplicatorConfigImpl();
+  public void init() {
+    security.runAsAdmin(
+        () -> {
+          try {
+            security.runWithSubjectOrElevate(
+                () -> {
+                  getAllConfigs()
```
clockard: I am wondering if we need some kind of retry logic in here for when we come up before the catalog is available.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/OldSite.java
```diff
@@ -0,0 +1,66 @@
+package org.codice.ditto.replication.api.impl.legacy;
+
+import java.util.Map;
+import org.codice.ditto.replication.api.data.ReplicationSite;
+import org.codice.ditto.replication.api.impl.data.AbstractPersistable;
+
+/** OldSite is a class that we can use to retrieve 0.2.0 version sites from the persistent store */
+public class OldSite extends AbstractPersistable implements ReplicationSite {
```
kcover: I had to create this class to access the old sites using the persistent store. It's the same as ReplicationSiteImpl aside from the PERSISTENCE_TYPE.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
```diff
@@ -54,97 +54,80 @@
...
+    this.security = security;
   }
 
-  public ReplicatorConfig createConfig() {
-    return new ReplicatorConfigImpl();
+  public void init() {
+    Failsafe.with(
+            new RetryPolicy()
+                .withBackoff(3, 60, TimeUnit.SECONDS)
+                .withMaxDuration(10, TimeUnit.MINUTES))
```
paouelle: ❓ Why are we stopping at 10 minutes?  Shouldn't we be continuing until it works?
❓ Should we not be re-trying for errors other than when Solr is not available?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
```diff
@@ -54,97 +54,80 @@
...
+        .onFailedAttempt(
+            throwable -> LOGGER.debug("failed to upgrade legacy configs, will retry", throwable))
+        .onFailure(
+            throwable ->
+                LOGGER.debug("failed to upgrade legacy configs, restart to try again.", throwable))
+        .run(this::convertConfigs);
```
paouelle: ❓ How would Failsafe detect errors and retry? From the `convertConfigs()` method, it seems to be returning successfully in all situations.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
```diff
@@ -54,97 +54,80 @@
...
-            "Error updating replication configuration for " + replicationConfig.getName());
-      }
-    }
-  }
-
-  public void removeConfig(ReplicatorConfig replicationConfig) {
+  void removeConfig(ReplicatorConfig replicationConfig) {
     try {
       framework.delete(new DeleteRequestImpl(replicationConfig.getId()));
     } catch (IngestException | SourceUnavailableException e) {
```
paouelle: ❓ Should we be handling `SourceUnavailableException` differently to let Failsafe above retry on these? Maybe a new `ReplicationSourceUnavailableException` or something similar extending `ReplicationException`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
```diff
@@ -154,104 +137,36 @@ public void removeConfig(ReplicatorConfig replicationConfig) {
...
-        config.getFailureRetryCount(),
-        DEFAULT_FAILURE_RETRY_COUNT);
-    helper.setIfPresentOrDefault(mcard, ReplicationConfig.VERSION, config.getVersion(), 1);
-    helper.setIfPresentOrDefault(mcard, ReplicationConfig.SUSPEND, config.isSuspended(), false);
-
-    return mcard;
-  }
-
   @VisibleForTesting
   ReplicatorConfig getConfigFromMetacard(Metacard mcard) {
```
paouelle: ✏️ Annotate method with `@Nullable`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
```diff
@@ -272,6 +187,27 @@ ReplicatorConfig getConfigFromMetacard(Metacard mcard) {
     return config;
   }
 
+  private ReplicatorConfigImpl setSourceAndDestinationWithOldUrl(
```
paouelle: ✏️ Annotate method with `@Nullable`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
```diff
@@ -54,97 +54,80 @@
...
+                "Error occured while trying to transform catalog saved metacards to persistent store saved configs. Legacy configs may not be accessible.",
+                e);
+          }
+          return null;
+        });
   }
 
-  public List<ReplicatorConfig> getAllConfigs() {
+  List<ReplicatorConfig> getAllConfigs() {
     return helper.getTypeForFilter(
```
paouelle: :question: Is there a way here to differentiate errors from unavailability to reach the data?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/LegacyDataConverter.java
```diff
@@ -35,16 +35,18 @@
...
 import org.codice.ditto.replication.api.impl.data.ReplicatorConfigImpl;
+import org.codice.ditto.replication.api.impl.persistence.ReplicationPersistentStore;
 import org.codice.ditto.replication.api.mcard.ReplicationConfig;
 import org.codice.ditto.replication.api.persistence.ReplicatorConfigManager;
 import org.codice.ditto.replication.api.persistence.SiteManager;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class MetacardConfigLoader {
+public class LegacyDataConverter {
```
paouelle: ✏️ Missing class documentation

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/OldSite.java
paouelle: ❓ Can't it extend the current site? Or could the `fromMap()` method of the `Site` class be able to handle the old version and the new one?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/LegacyDataConverter.java
```diff
@@ -0,0 +1,248 @@
...
+
+  List<ReplicatorConfig> getAllConfigs() {
+    return helper.getTypeForFilter(
+        filterBuilder.attribute(Metacard.TAGS).is().like().text(ReplicationConfig.METACARD_TAG),
+        this::getConfigFromMetacard);
+  }
+
+  void removeConfig(ReplicatorConfig replicationConfig) {
+    try {
+      framework.delete(new DeleteRequestImpl(replicationConfig.getId()));
```
peterhuffer: ❓ Should we go straight to the provider to avoid versioning off the config?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/LegacyDataConverter.java
```diff
@@ -0,0 +1,248 @@
...
+  private void convertConfigs() {
+    // convert sites first
+    persistentStore.objects(OldSite.class).forEach(this::convert020Site);
+
+    security.runAsAdmin(
+        () -> {
+          try {
+            security.runWithSubjectOrElevate(
+                () -> {
+                  getAllConfigs()
```
peterhuffer: ❓ Looks like `getConfigFromMetacard` can return null, should we filter nulls out here?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
kcover: Based on the examples I've seen this seems to retry when any exception is thrown. When running unit tests at one point I was getting a lot of NPEs and the unit tests that ran through failsafe would hang for 10 minutes so I beleive they were hanging because they kept seeing NPEs as well.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
kcover: If solr doesn't come up 10 minutes after this bundle does than I didn't think It'd be likely to come up at all but I guess I don't see a good reason for not just continuing to retry every minute. To answer your 3rd question I think this retries for any error that occurs, not just solr.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
paouelle: That being said. On one side, Failsafe would, as you stated, retry on all failures. There are some type of exceptions (i.e. `Throwable`) that should never be ignored. Those are the `VirtualMachineError`. No one should ignore those and continue running as they should be left to bubble out. Failsafe doesn't do that properly and as such, it should not be used with its default configuration. Secondly, as stated below, your `convertConfigs()` method catch exceptions and return successfully with `null` which would let Failsafe believe you were successful.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
paouelle: On the contrary, the system is designed to continuously retry to connect to Solr in the background without having the admin come in and be forced to restart us. Remember that we support not just the Solr instance we start from the `ddf` script but we also support Solr cloud or an admin managed Solr instance. You should therefore continue retrying. That is if the error is Solr not being available versus another error for which retry will never ever help.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/LegacyDataConverter.java
kcover: I'm not sure if we care about that. we were using the framework before so I assume this would still be what we want to use.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/OldSite.java
kcover: I didn't know how that would work with trying to override a static field but I guess I can. I think the from map needs to stay the same however. Our new site's are always going to have a version and we need to read it while our old ones will always be null. we could do a null check in the Site class but I'd rather code the class we're generally going to be using for this one specific case.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
kcover: I'm not sure what specifically causes an IngestException so I'm not sure whether it means we should try again. I think we should leave it for now and try again on either exception just to be safe. I don't think we need to log and rethrow the exception however

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/OldSite.java
paouelle: Then how about just extending the class and leave the type intact override the fromMap() to fill-in the version field as expected by the base class and then call super.fromMap()? That way, you don't pollute the base class and you don't end up creating a bran new type.

replication-api-impl/pom.xml
```diff
@@ -335,7 +335,7 @@
                                         <limit implementation="org.codice.jacoco.LenientLimit">
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.39</minimum>
+                                            <minimum>0.35</minimum>
```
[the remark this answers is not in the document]
kcover: don't cry too much, this is from removing tested code, not adding untested code.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/MetacardConfigLoader.java
paouelle: One of the problem I see is that if you cannot delete it, than you will be continuously trying to re-upgrade it and potentially override what was already updated by users since the first time it was upgraded. Maybe, we should add a check to see if a particular object already exist in the new repo before attempting to override it.

`SourceUnavailableException` definitely sounds like one we should retry on.

clockard:
Verified conversion from 0.2.0 to 0.2.1 works correctly ✅

PR #43, opened 2019-03-19
Upgrade DDF version

peterhuffer:
#### What does this PR do?
- Upgrade DDF version to 2.13.7-SNAPSHOT

@clockard @paouelle @kcover 

Full build.

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
         <releases.repository.url />
         <admin-query.version>1.2.7</admin-query.version>
         <ddf.support.version>2.3.12</ddf.support.version>
-        <ddf.version>2.13.6</ddf.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -28,7 +28,7 @@
         <releases.repository.url />
         <admin-query.version>1.2.7</admin-query.version>
```
clockard: Are we also going to need to bump the admin-query version?

peterhuffer: I think it will pull in a version range of DDF. Having trouble checking right now though cause solr is broken, but will check before merging.

peterhuffer: Confirmed no 2.13.6 DDF bundles are started from admin-query.

PR #44, opened 2019-03-19
REP-42 Update status while processing

### LINKED ISSUE #42: Update replication status as items are processed
**Describe the task**
Right now the status of a replication job is only reported once it is finished. This task is to make it so the status gets updated as each individual piece of the job is completed so anyone looking at the status can see the progress.

clockard:
#### What does this PR do?
Updates the status (item count and bytes transferred) while the replication is running
Ingest a fairly large amount of data such that it will take a minute or so to replicate. Setup a replication job to replicate the ingested data. Observe that the item count and bytes transferred are updated as the job is running before it is completed
Fixes #42

### DIFF
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationStatusImpl.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationStatusImpl.java
@@ -1,0 +1,237 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ditto.replication.api.impl.data;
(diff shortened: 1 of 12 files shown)


replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationStatusImpl.java
```diff
@@ -0,0 +1,237 @@
...
+ */
+package org.codice.ditto.replication.api.impl.data;
+
+import java.util.Date;
+import java.util.UUID;
+import javax.annotation.Nullable;
+import org.codice.ditto.replication.api.ReplicationStatus;
+import org.codice.ditto.replication.api.Status;
+
+public class ReplicationStatusImpl implements ReplicationStatus {
```
paouelle: ✏️ Missing class documentation

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicationStatus.java
```diff
@@ -14,169 +14,141 @@
...
+  Date getLastSuccess();
+
+  void setLastSuccess(@Nullable Date lastSuccess);
+
+  /**
+   * Gets the runtime duration of the referenced configuration in seconds
+   *
+   * @return
+   */
+  long getDuration();
```
paouelle: ❓ Why set it to seconds? Why not keep it in milliseconds?

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicationStatus.java
```diff
@@ -14,169 +14,141 @@
...
+  /** Sets the start time to now */
+  void markStartTime();
+
+  /**
+   * Get the Date of the last time the referenced configuration was run
+   *
+   * @return
+   */
+  @Nullable
+  Date getLastRun();
```
paouelle: ✏️ Using java.time.Instant might be preferable over java.util.Date

PR #51, opened 2019-03-27
REP-50 Set import version range on ddf.security.encryption

### LINKED ISSUE #50: Missing dependency error when deploying replication on older versions of DDF
**Describe the bug**
When deploying the current replication 0.2.1-SNAPSHOT replication on DDF v2.13.6 or earlier a missing dependency error is thrown for ddf.security.encryption. This happened when replication was updated to ddf 2.13.7-SNAPSHOT where that dependency was fixed to export itself as the project version rather than 1.0.0. 
Breaking change is in https://github.com/codice/ddf/commit/2247de634cc61dd9a90fe6bea0597cca7003c9aa
**Affects version**
0.2.1-SNAPSHOT
**To Reproduce**
Build current replication master and deploy the resulting kar to a ddf-2.13.6

**Expected behavior**
Current replication master should deploy to ddf-2.13.6 without any missing dependencies

clockard:
#### What does this PR do?
Fixes the ddf.security.encryption dependency issue
@kcover @peterhuffer @paouelle 
Deploy the kar to an instance of ddf-2.13.6 and verify there are no missing dependency errors

Fixes #50

### DIFF
--- a/replication-api-impl/pom.xml
+++ b/replication-api-impl/pom.xml
@@ -284,6 +284,7 @@
                         <Import-Package>
                             !org.jvnet.staxex,
                             ddf.security.permission,
+                            ddf.security.encryption;version="[1.0.0, 3.0.0]",
                             *
                         </Import-Package>
                         <Export-Package>


replication-api-impl/pom.xml
```diff
@@ -284,6 +284,7 @@
                         <Import-Package>
                             !org.jvnet.staxex,
                             ddf.security.permission,
+                            ddf.security.encryption;version="[1.0.0, 3.0.0]",
```
peterhuffer: ❓ Why not `(2.13, 3)`?

clockard: Because 2.13.6 and earlier export that bundle as version 1.0.0

PR #52, opened 2019-03-27
[WIP] REP-45 Add option to delete replication data through UI

peterhuffer:
#### What does this PR do?
More tests coming.

- Fixes an issue with erroneous data transfer if 2 replications had overlapping filters. The fix was making sure a replication item always has a unique id (changed from metacard id)
- Fixes an issue with karaf configuration delete command not cleaning up replication items.
- Fixes an issue in the UI where clicking the save configuration button twice quickly would save 2 configurations with the same name
- Adds option to delete replicated data when deleting a configuration. Only data replicated from another node to the local node will be deleted.
- Add a scheduled deleter that checks for deleted configurations and cleans them up (runs every minute)
- Add checks for deleting orphaned replication items (replication items without corresponding configurations or metacards).

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?
@clockard @paouelle @kcover @mcalcote 

#### How should this be tested? (List steps with links to updated documentation)
- Replicate and verify deleting data along with the replication.
- Replicate and delete a config (preserver data). Delete a metacard and verify its corresponding replication item entry gets cleaned up.

#### What are the relevant tickets?
Fixes #45 

#### Screenshots (if appropriate)
image

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

PR #53, opened 2019-03-27
Various bug fixes

peterhuffer:
#### What does this PR do?
Changes pulled from https://github.com/connexta/replication/pull/52

- Fixes an issue where a request that was running could have an immediate follow up in the pending requests and the pending request run before the `Last Run` on history was set, meaning all the metacards would be queried again.
- Fixes an issue with karaf configuration delete command not cleaning up replication items.
- Ignore case when creating new replication configs and nodes.

@clockard @kcover @paouelle 

Run `log:set DEBUG org.codice.ditto.replication.api.impl`

- Perform a replication and verify the same replicator request is not added to pending requests if it is running (look at logs).
- Confirm `replication:delete -c -h -m` deletes replication items in solr as well as the config and all metacards/data.

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/persist/DeleteReplication.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/persist/DeleteReplication.java
@@ -22,6 +22,7 @@ import org.codice.ddf.admin.api.fields.FunctionField;
 import org.codice.ddf.admin.common.fields.base.BaseFunctionField;
 import org.codice.ddf.admin.common.fields.base.scalar.BooleanField;
 import org.codice.ddf.admin.common.fields.common.PidField;
+import org.codice.ditto.replication.admin.query.ReplicationMessages;
 import org.codice.ditto.replication.admin.query.ReplicationUtils;
 
 public class DeleteReplication extends BaseFunctionField<BooleanField> {
@@ -51,6 +52,18 @@ public class DeleteReplication extends BaseFunctionField<BooleanField> {
   }
 
   @Override
+  public void validate() {
+    super.validate();
+    if (containsErrorMsgs()) {
+      return;
+    }
+
+    if (!replicationUtils.configExists(id.getValue())) {
+      addErrorMessage(ReplicationMessages.configDoesNotExist());
+    }
+  }
+
+  @Override
   public BooleanField getReturnType() {
     return RETURN_TYPE;
   }
@@ -67,6 +80,6 @@ public class DeleteReplication extends BaseFunctionField<BooleanField> {
 
   @Override
   public Set<String> getFunctionErrorCodes() {
-    return ImmutableSet.of();
+    return ImmutableSet.of(ReplicationMessages.CONFIG_DOES_NOT_EXIST);
   }
 }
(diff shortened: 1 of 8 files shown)


peterhuffer:
build now

peterhuffer:
build now

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorConfigManagerImpl.java
```diff
@@ -43,4 +45,16 @@ public void save(ReplicatorConfig replicatorConfig) {
...
+  public boolean configExists(String configId) {
+    try {
+      persistentStore.get(ReplicatorConfigImpl.class, configId);
+    } catch (ReplicationPersistenceException e) {
+      return true;
+    } catch (NotFoundException e) {
+      return false;
+    }
+    return true;
+  }
```
kcover: ✏️ could just use the get(id) method and if it throws the NotFoundException you know the config doesn't exist. Also, it seems a little misleading to return true when we get a ReplicationPersistenceException.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/persist/DeleteReplication.java
```diff
@@ -50,6 +51,18 @@ public BooleanField performFunction() {
...
+  public void validate() {
+    super.validate();
+    if (containsErrorMsgs()) {
+      return;
+    }
+
+    if (!replicationUtils.configExists(id.getValue())) {
+      addErrorMessage(ReplicationMessages.configDoesNotExist());
+    }
+  }
```
kcover: replicationUtils.deleteConfig(id) could be changed to just let the NotFoundException bubble up. Then in the performFunction() method you could just return null with the error message if you catch that exception. At least I think you can do that with the graphql endpoint.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorConfigManagerImpl.java
clockard: I would agree that we shouldn't return true on a ReplicationPersistenceException

kcover:
Successful hero with the test steps above.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorConfigManagerImpl.java
peterhuffer: I changed to return false for now, but I think we will want to throw some sort of error and handle on the front end if we cannot access the persistence store.

peterhuffer:
build now

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorConfigManagerImpl.java
clockard: I think that is exactly right. We want to throw an exception up if there is an error instead of returning true/false.

PR #54, opened 2019-03-27
Fix Jenkinsfile default parameter values

clockard:
#### What does this PR do?
What the title says
@kcover @peterhuffer 
Already tested on jenkins environment


clockard:
I don't expect this to pass the build right now because of a known test failure. If it fails the build for that reason that should be good enough for a merge

clockard:
Build failed for test case but jenkins file worked as expected

Issue #58, opened 2019-03-29
Remove deprecated features


kcover:
Remove code that is no longer needed. This includes:
- the CLI code
- code generating a local node
- any code refering to a local catalog store
- legacy data conversion logic

PR #59, opened 2019-03-29
REP-58 Remove CLI

kcover:
#### What does this PR do?
Removes the replication-command component as we're not going to be using
it going forward.

#### Who is reviewing it 
@peterhuffer @paouelle @clockard 

CI build should be good enough I think

Work for: #58 

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests


peterhuffer:
I'm curious on thoughts for keeping the delete and run commands? Since both deleting and running configurations are based on polling in the back-end, these commands allow for faster testing. The run command functionality can eventually be added to the UI then removed.

kcover:
Is there any reason why we can't just add a replication to the run queue as soon as it's created? I think that would be an easy change and would remove the need for a run command.

peterhuffer:
That wouldn't allow an already existing config to be run at any time. They would have to delete the config and remake it.

PR #61, opened 2019-03-29
REP-58  Remove LocalCatalogResourceStore and local node generation 

kcover:
#### What does this PR do?
Removes references to a local node since we won't be tied to a node going
forward.

install replication and verify no local node is created. Then run a quick replication and verify it was successful.

Work for: #58

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests


peterhuffer:
Successful hero ✅

PR #62, opened 2019-03-29
REP-58 Remove catalog coupled code

kcover:
#### What does this PR do?
Removes the LegacyDataConverter, replication config metacard attributes,
and other ddf coupled code that we don't need anymore. Also refactors the ReplicatorHistory to be a DataManager so that it uses the persistent store rather than the catalog.

Install replication on a DDF where you can access solr
Create a replication config and let it run
Check solr to make sure the history was created correctly
let the config run again
Check solr to make sure the history was updated correctly

Work for: #58 

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests


kcover:
I just realized that this will need #59 to be removed first as it won't build currently

replication-api-impl/pom.xml
```diff
@@ -331,12 +331,12 @@
                                         <limit implementation="org.codice.jacoco.LenientLimit">
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.54</minimum>
+                                            <minimum>0.50</minimum>
```
paouelle: 😢

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorHistoryManagerImpl.java
```diff
@@ -102,34 +103,38 @@ public void init() {
...
             .is()
             .equalTo()
             .text(ReplicationHistory.METACARD_TAG),
-        this::getStatusFromMetacard);
+        this::getStatusFromMetacard).stream();
   }
 
   @Override
-  public List<ReplicationStatus> getReplicationEvents(String replicatorid) {
+  public Stream<ReplicationStatus> getByReplicatorId(String replicatorid) {
```
paouelle: ✏️ methods that returned `Stream<>` should not be prefixed with `get` by Java convention

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
```diff
@@ -201,7 +197,7 @@ private Filter buildFilter() {
     final List<Filter> failureFilters = createFailedMetacardFilters();
     final ReplicationStatus lastSuccessfulRun =
         history
-            .getReplicationEvents(config.getName())
+            .getByReplicatorId(config.getName())
```
paouelle: ❗️ ❓ Should we not be cross-referencing objects using their internal unmodifiable ids as opposed to names specified by users? Names can be changed, not ids.

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicatorHistoryManager.java
```diff
@@ -13,40 +13,23 @@
...
-   */
-  void addReplicationEvent(ReplicationStatus replicationStatus);
-
-  /**
-   * Remove a replication event from the history
-   *
-   * @param replicationStatus replication event to remove
-   */
-  void removeReplicationEvent(ReplicationStatus replicationStatus);
+  Stream<ReplicationStatus> getByReplicatorId(String replicatorid);
```
paouelle: See previous comment about naming conventions and consistency

replication-api/src/main/java/org/codice/ditto/replication/api/ReplicatorHistoryManager.java
```diff
@@ -13,40 +13,23 @@
...
-   *
-   * @return List of all replication events
-   */
-  List<ReplicationStatus> getReplicationEvents();
+public interface ReplicatorHistoryManager extends DataManager<ReplicationStatus> {
 
   /**
    * Get all replication events for a given replication configuration
    *
    * @param replicatorid replication configuration id
```
paouelle: ❓ Is this a replication id or a config name?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/mcard/ReplicationHistoryAttributes.java
```diff
@@ -123,7 +122,13 @@
             false /* multivalued */,
             BasicTypes.STRING_TYPE));
     descriptors.add(
-        new ReplicationConfigAttributes().getAttributeDescriptor(ReplicationConfig.NAME));
+        new AttributeDescriptorImpl(
```
clockard: We shouldn't need this file at all anymore since we are not storing anything as metacards right?

replication-api-impl/src/main/resources/OSGI-INF/blueprint/blueprint.xml
```diff
@@ -106,19 +94,19 @@
 
     <service ref="configManager" interface="org.codice.ditto.replication.api.persistence.ReplicatorConfigManager"/>
 
-    <bean id="historyService" class="org.codice.ditto.replication.api.impl.ReplicatorHistoryImpl" init-method="init">
+    <bean id="historyService" class="org.codice.ditto.replication.api.impl.persistence.ReplicatorHistoryManagerImpl" init-method="init">
```
clockard: These are now the wrong arguments for this class. After this pr there shouldn't be any references to the catalog.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -159,7 +160,20 @@ private ReplicationField getReplicationFieldForConfig(ReplicatorConfig config) {
...
     field.filter(config.getFilter());
     field.biDirectional(config.isBidirectional());
-    List<ReplicationStatus> statusList = history.getReplicationEvents(config.getName());
+    field.suspended(config.isSuspended());
+    field = setStatusRelatedFields(field, config);
+
+    return field;
+  }
+
+  private ReplicationField setStatusRelatedFields(ReplicationField field, ReplicatorConfig config) {
```
paouelle: ❓ Why return field here? I would see the usefulness if above it was used for chaining but it is not.

replication-api-impl/pom.xml
```diff
@@ -331,12 +331,12 @@
                                         <limit implementation="org.codice.jacoco.LenientLimit">
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.54</minimum>
+                                            <minimum>0.43</minimum>
```
paouelle: ✏️ Definitely not going in the right direction

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
```diff
@@ -110,8 +110,9 @@ void scheduleReplication() {
             .collect(Collectors.toList());
     try {
       for (ReplicatorConfig config : configsToSchedule) {
-        replicator.submitSyncRequest(
-            new SyncRequestImpl(config, new ReplicationStatusImpl(config.getName())));
+        ReplicationStatusImpl status = new ReplicationStatusImpl();
```
paouelle: I personally preferred the simpleness of the previous code.
I like that we are correlating with config ids and not arbitrary names now.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
```diff
@@ -214,12 +204,16 @@ private Filter buildFilter() {
...
-      long time = lastSuccessfulRun.getStartTime().getTime();
-      if (lastSuccessfulRun.getLastSuccess() != null) {
-        time = lastSuccessfulRun.getLastSuccess().getTime();
-      }
-      Date timeStamp = new Date(time - 1000);
+    long timeOfLastSuccessfulRun = 0;
+    try {
+      timeOfLastSuccessfulRun =
+          history.getByReplicatorId(config.getId()).getLastSuccess().getTime();
+    } catch (NotFoundException | NullPointerException e) {
```
paouelle: ❗️ We shouldn't rely on our own code to generate `NullPointerException`; this is bad practice. If we know something above can be `null`, we should check for it manually there.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
```diff
@@ -214,12 +204,16 @@ private Filter buildFilter() {
...
+      // config was never run or never successful
+    }
+
+    if (timeOfLastSuccessfulRun > 0) {
+      Date timeStamp = new Date(timeOfLastSuccessfulRun - 1000);
```
paouelle: ❓ Why `- 1000`? THe reason should be documented here for the next person looking at the code.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationStatusImpl.java
```diff
@@ -47,24 +71,21 @@
...
-  public String getId() {
-    return id;
+  public String getReplicatorId() {
+    return replicatorId;
   }
 
   @Override
-  public String getReplicatorName() {
-    return replicatorName;
+  public void setReplicatorId(String replicatorName) {
```
paouelle: ✏️ Should change the parameter name

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorHistoryManagerImpl.java
```diff
@@ -0,0 +1,107 @@
...
+  public Stream<ReplicationStatus> objects() {
+    return persistentStore.objects(ReplicationStatusImpl.class).map(ReplicationStatus.class::cast);
+  }
+
+  @Override
+  public ReplicationStatus getByReplicatorId(String id) {
+    return objects()
+        .filter(status -> status.getReplicatorId().equals(id))
+        .findFirst()
+        .orElseThrow(NotFoundException::new);
```
paouelle: 👍 Nice. However, a little bit more info on the exception would be a good idea to indicate what was not found

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorHistoryManagerImpl.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  @Override
+  public void save(ReplicationStatus status) {
+    if (!(status instanceof ReplicationStatusImpl)) {
+      throw new IllegalArgumentException(
+          "Expected a ReplicationStatusImpl but got a " + status.getClass().getSimpleName());
+    } else {
```
paouelle: ✏️ No need for the `else` here; just fall-through since the `if` throws an exception

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorHistoryManagerImpl.java
```diff
@@ -0,0 +1,107 @@
...
+
+      try {
+        ReplicationStatus previous = getByReplicatorId(status.getReplicatorId());
+        ReplicationStatus updatedStatus = addStats(previous, status);
```
paouelle: ❗️ This is going to be problematic as we are compounding stats info ourselves here and if multiple `saves` happens on multiple threads or in the future on multiple processes, we will be overwriting info saved by another and thus loosing stats. For cloud based development, you either want to save individual entries and sum them up when you want to display or get the database to help you by having it add or update data.

Keeping stats like this in a single row in the DB is never easy to do unless you can guarantee that only one is ever reading the previous and updating it before anybody can read it. In addition, it does require that every time you need to update, you are forced to load the previous version; thus making 2 DB calls. Easier to penalize the read than the write.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorHistoryManagerImpl.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  @Override
+  public void remove(String id) {
+    persistentStore.delete(ReplicationStatusImpl.class, id);
+  }
+
+  @VisibleForTesting
+  ReplicationStatus addStats(ReplicationStatus base, ReplicationStatus status) {
```
paouelle: ❓ Is there a need to return `base` back?

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/SyncHelperTest.java
```diff
@@ -64,14 +65,15 @@
...
-  @Mock ReplicatorHistory history;
+  @Mock ReplicatorHistoryManager history;
 
   ReplicationStatus status;
 
   @Before
   public void setUp() throws Exception {
     FilterBuilder builder = new GeotoolsFilterBuilder();
-    status = new ReplicationStatusImpl("test");
+    status = new ReplicationStatusImpl();
```
paouelle: We could have kept the other constructor as an overloaded one. There was no harm in it and did simplify the code.

replication-api/src/main/java/org/codice/ditto/replication/api/data/ReplicationStatus.java
```diff
@@ -30,11 +31,13 @@
   String getId();
 
   /**
-   * Get the name of the associated replication configuration
+   * Get the ID of the associated replication configuration
```
paouelle: ✏️ Maybe clarify that it gets the unique identifier for the configuration object representing the replicator. Something like that.

Should be `Gets` as opposed to `Get`

replication-api/src/main/java/org/codice/ditto/replication/api/data/ReplicationStatus.java
```diff
@@ -30,11 +31,13 @@
...
    *
-   * @return The replicator name
+   * @return The replicator ID
    */
-  String getReplicatorName();
+  String getReplicatorId();
+
+  void setReplicatorId(String replicatorId);
```
paouelle: ✏️ Missing javadoc

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicatorHistoryManager.java
```diff
@@ -0,0 +1,32 @@
...
+package org.codice.ditto.replication.api.persistence;
+
+import org.codice.ditto.replication.api.ReplicationPersistenceException;
+import org.codice.ditto.replication.api.data.ReplicationStatus;
+
+/** Persistence interface for storing, retrieving, and deleting history items */
+public interface ReplicatorHistoryManager extends DataManager<ReplicationStatus> {
+
+  /**
+   * Get the replication history for a given replication configuration
```
paouelle: ✏️ `Gets`

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicatorHistoryManager.java
```diff
@@ -0,0 +1,32 @@
...
+   *
+   * @param replicatorId replication configuration id
```
paouelle: ✏️ This description for `replicatorId` is clear

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/ReplicatorImplTest.java
```diff
@@ -28,18 +28,14 @@
...
 import java.util.concurrent.ExecutorService;
 import org.codice.ddf.security.common.Security;
-import org.codice.ditto.replication.api.ReplicationPersistentStore;
-import org.codice.ditto.replication.api.ReplicationStatus;
-import org.codice.ditto.replication.api.ReplicationStore;
-import org.codice.ditto.replication.api.ReplicatorHistory;
-import org.codice.ditto.replication.api.ReplicatorStoreFactory;
-import org.codice.ditto.replication.api.Status;
-import org.codice.ditto.replication.api.SyncRequest;
+import org.codice.ditto.replication.api.*;
```
peterhuffer: ✏️ Star import.

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorHistoryManagerImplTest.java
```diff
@@ -0,0 +1,196 @@
...
+  @Test(expected = NotFoundException.class)
+  public void getByReplicatorIdNotFoundException() {
+    ReplicationStatusImpl status = new ReplicationStatusImpl();
+    status.setReplicatorId("test");
+    when(persistentStore.objects(eq(ReplicationStatusImpl.class))).thenReturn(Stream.empty());
+    history.getByReplicatorId("test");
+  }
+
+  @Test
+  public void save() {
```
peterhuffer: ❓ Is this test needed since we have the addStatus test? `addStatus` can be private and tested indirectly through the save method. I think most the tests below should be going through the save method since it will ensure that `save` is doing what we expect.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
kcover: Imo it conveys more clearly that we are modifying the field. Even better would be to have the method on the field class so it would read field.setStatusRelatedFields() but then I would have to pass in the config, history, and replicator.

paouelle: ok. but then, from this method prototype, one would need to start using the returned field as you wouldn't know if a new one is returned or the same one.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
paouelle: my point 😉

replication-api-impl/pom.xml
kcover: The code that I've added is 100% covered but this is still going down because the code that we had before was still mostly tested but there was a lot more of it. I could add tests for random classes in this PR but I think that would be better left for another one. I want to increase our test coverage too, maybe we should bring it up at our next planning meeting.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
kcover: I find it a little easier to read this way and I miss the id constructor a little bit but this aligns more with the way that we initialize our other data objects.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationStatusImpl.java
```diff
@@ -47,24 +71,21 @@
...
-    this.id = UUID.randomUUID().toString();
-  }
+  /** 1 - initial version. */
+  public static final int CURRENT_VERSION = 1;
 
-  public ReplicationStatusImpl(String id, String replicatorName) {
-    this.id = id;
-    this.replicatorName = replicatorName;
+  public ReplicationStatusImpl() {
+    super.setVersion(CURRENT_VERSION);
```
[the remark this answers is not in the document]
kcover: The default constructor for AbstractPersistable automatically gets called here and that sets the ID.

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/persistence/ReplicatorHistoryManagerImplTest.java
kcover: the addStats method is now on the ReplicationStatus class and is tested seperately.

replication-api-impl/pom.xml
```diff
@@ -331,17 +331,17 @@
                                         <limit implementation="org.codice.jacoco.LenientLimit">
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.43</minimum>
+                                            <minimum>0.61</minimum>
```
paouelle: 🎉

clockard:
Heroed successfully ✅

PR #67, opened 2019-04-17
REP-63 Remove DDF dependencies from API

### LINKED ISSUE #63: Remove DDF CatalogStore from internal API
**Describe the task**
Remove DDF interfaces (CatalogStore, CreateRequests, etc) that are currently being exposed through the internal API.

**Additional context**
This is part of the move to deploying as a separate application outside of DDF.

peterhuffer:
#### What does this PR do?
- Removes DDF dependencies from API
- Querying after metadata modified date is now dependent upon the last processed Metadata's modified date rather than the last successful run,

@paouelle @kcover @clockard @mcalcote 

Setup a replication and test in order:

- Verify creating. Check replicated metacard's Origins attribute, the replicated tag is present, and that the product can be downloaded.
- Verify updating
- Verify deleting
- Verify another create
- Verify losing connection while replication then reconnecting and continuing replication
- Verify failing to transfer an item and then fixing the transfer issue (easiest with CDM in place without UrlResourceReader properly configured)

Fixes #63 

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests


adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
```diff
@@ -0,0 +1,569 @@
...
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  @Override
+  public void close() {
+    super.destroy(9001);
```
paouelle: 😕 `9001`?

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapterFactory.java
```diff
@@ -77,8 +84,12 @@
...
 
-  public ReplicationStore createReplicatorStore(URL url) {
+  public DdfNodeAdapterFactory(DdfRestClientFactory ddfRestClientFactory) {
+    this.ddfRestClientFactory = ddfRestClientFactory;
+  }
 
+  @Override
+  public NodeAdapter create(URL url) {
     CswSourceConfiguration cswConfiguration = new CswSourceConfiguration(encryptionService);
     cswConfiguration.setCswUrl(url.toString() + "/services/csw");
```
paouelle: ❓ Are we guaranteed the root context path is always `/services`? I thought it was customizable?

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapterFactory.java
```diff
@@ -96,23 +107,29 @@ public ReplicationStore createReplicatorStore(URL url) {
...
+
+    return ddfNodeAdapter;
+  }
+
+  @Override
+  public NodeAdapterType getType() {
+    return NodeAdapterType.DDF;
   }
 
   public void setBundleContext(BundleContext bundleContext) {
```
paouelle: ❗️ Caching a bundle context tends to create problem as it can get invalidated when the bundle is refreshed or otherwise. It is safer to get it back each time you need it.

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfRestClientFactory.java
```diff
@@ -0,0 +1,151 @@
...
+import org.codice.ddf.cxf.client.SecureCxfClientFactory;
+import org.codice.ddf.endpoints.rest.RESTService;
+import org.codice.ddf.security.common.Security;
+
+/** Factory for creating a basic client capable of posting resources to the DDF REST endpoint. */
+public class DdfRestClientFactory {
+
+  private static final String CONTENT_DISPOSITION = "Content-Disposition";
+
+  private static final String DEFAULT_REST_ENDPOINT = "/services/catalog";
```
paouelle: ✏️  Hardcoding the root context path. I though it was customizable on DDF

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfRestClientFactory.java
```diff
@@ -0,0 +1,151 @@
...
+import org.apache.cxf.jaxrs.ext.multipart.ContentDisposition;
+import org.apache.cxf.jaxrs.ext.multipart.MultipartBody;
+import org.apache.cxf.jaxrs.impl.MetadataMap;
+import org.codice.ddf.cxf.client.ClientFactoryFactory;
+import org.codice.ddf.cxf.client.SecureCxfClientFactory;
+import org.codice.ddf.endpoints.rest.RESTService;
+import org.codice.ddf.security.common.Security;
+
+/** Factory for creating a basic client capable of posting resources to the DDF REST endpoint. */
+public class DdfRestClientFactory {
```
paouelle: ❗️ Could definitely use a bit more javadoc throughout

replication-api-impl/pom.xml
```diff
@@ -331,17 +148,17 @@
                                         <limit implementation="org.codice.jacoco.LenientLimit">
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.61</minimum>
+                                            <minimum>0.00</minimum>
```
paouelle: ❗️ ❓ Why?
If we are planning on managing the code that is brought in then it must be tested at the level required for quality.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/NodeAdapters.java
```diff
@@ -0,0 +1,26 @@
+package org.codice.ditto.replication.api.impl;
+
+import java.util.List;
+import org.codice.ditto.replication.api.NodeAdapterFactory;
+import org.codice.ditto.replication.api.NodeAdapterType;
+
+public class NodeAdapters {
```
paouelle: ❗️ Javadoc. This comments applies to a lot of places. All classes and public methods should always be javadoc

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorImpl.java
```diff
@@ -136,15 +121,15 @@ public void init() {
                 executeSyncRequest(syncRequest);
               }
             } catch (InterruptedException e) {
-              LOGGER.trace("InterruptedException in executor. This is expected during shutdown.");
+              LOGGER.debug("InterruptedException in executor. This is expected during shutdown.");
```
paouelle: ✏️ This log is not too useful and should be left at trace if not removed.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorImpl.java
```diff
@@ -136,15 +121,15 @@ public void init() {
...
               Thread.currentThread().interrupt();
               break;
             }
           }
         });
 
-    LOGGER.trace(
-        "Successfully configured the single-thread scheduler to execute sync requests from the queue");
+    LOGGER.info(
```
paouelle: ❓ info? This seems like developer's info and not customers so debug would be more appropriate

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/Syncer.java
```diff
@@ -0,0 +1,342 @@
...
+
+    private final String sourceName;
+
+    private final String destinationName;
+
+    private final ReplicatorConfig replicatorConfig;
+
+    private ReplicationStatus replicationStatus;
+
+    private boolean canceled = false;
```
paouelle: ❗️ This attribute seems to be designed to be accessed from multiple threads and as such all access should either be protected by synchronization or be made volatile based on careful considerations.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/Syncer.java
```diff
@@ -0,0 +1,342 @@
...
+
+    private final NodeAdapter destination;
+
+    private final String sourceName;
+
+    private final String destinationName;
+
+    private final ReplicatorConfig replicatorConfig;
+
+    private ReplicationStatus replicationStatus;
```
paouelle: ❗️ Because it will be accessed by multiple threads, special considerations applies to both the reference and the object itself. If the reference is never changed then we should make it final. All access to the object should be protected by synchronization or the object itself should be designed to be thread safe.

replication-api/src/main/java/org/codice/ditto/replication/api/NodeAdapter.java
```diff
@@ -0,0 +1,94 @@
...
+   * @param createRequest request containing {@link Metadata} to create.
+   * @return {@code true} if the creation was successful, {@code false} otherwise
+   */
+  boolean createRequest(CreateRequest createRequest);
+
+  /**
+   * Updates {@link Metadata} on this {@code NodeAdapter}.
+   *
+   * @param updateRequest request containing {@link Metadata} to update.
+   * @return {@code true} if the update was successful, {@code false} otherwise
```
paouelle: ❓ Why return true/false as opposed to throwing more meaningful exceptions in case of failures?

replication-api/src/main/java/org/codice/ditto/replication/api/Status.java
```diff
@@ -13,14 +13,44 @@
  */
 package org.codice.ditto.replication.api;
 
-/** The states that a request can be in. */
+/** The states that a {@link SyncRequest} can be in. */
 public enum Status {
+
```
paouelle: 🎉 Thank you!

replication-api/src/main/java/org/codice/ditto/replication/api/data/Metadata.java
```diff
@@ -0,0 +1,96 @@
...
+  Date getResourceModified();
+
+  /**
+   * @param resourceModified the {@link Date} the {@link Resource} represented by this metadata was
+   *     last modified
+   */
+  void setResourceModified(Date resourceModified);
+
+  /** @return a set of tags associated with the wrapped metadata */
+  Set<String> getTags();
```
paouelle: ❓ Should we consider not returning collections in interfaces like that to avoid either exposing internal structures or be forced to clone or wrap the internal collection int an unmodifable one?

replication-api/src/main/java/org/codice/ditto/replication/api/data/QueryResponse.java
```diff
@@ -0,0 +1,16 @@
...
+   * An iterable of {@link Metadata} returned by a {@link QueryRequest} sent to a {@link
+   * org.codice.ditto.replication.api.NodeAdapter}, which translates a node's specific metadata
+   * format into {@link Metadata}.
+   *
+   * <p>Considerations of paging when iterating should be taken into account in order to avoid
+   * memory issues.
+   *
+   * @return the iterable of {@link Metadata}
+   */
+  Iterable<Metadata> getMetadata();
```
paouelle: ✏️ Stream instead of iterators?

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicationItemManager.java
```diff
@@ -11,26 +11,28 @@
...
+import org.codice.ditto.replication.api.ReplicationItem;
 
-public interface ReplicationPersistentStore {
+public interface ReplicationItemManager {
 
   Optional<ReplicationItem> getItem(String id, String source, String destination);
 
-  List<ReplicationItem> getItemsForConfig(String configId, int startIndex, int pageSize)
-      throws PersistenceException;
+  /** @throws org.codice.ditto.replication.api.ReplicationPersistenceException */
```
paouelle: ❗️ Missing javadoc info

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorImpl.java
```diff
@@ -311,11 +288,11 @@ public void cancelSyncRequest(String configId) {
...
   }
 
-  private ReplicationStore getStoreForId(String siteId) {
-    ReplicationStore store;
+  private NodeAdapter getStoreForId(String siteId) {
+    NodeAdapter store;
     ReplicationSite site = siteManager.get(siteId);
     try {
-      store = replicatorStoreFactory.createReplicatorStore(new URL(site.getUrl()));
+      store = nodeAdapters.factoryFor(NodeAdapterType.DDF).create(new URL(site.getUrl()));
```
clockard: I think we want the factory to figure out which kind of adapter we need. Soon when we have another type of adapter we will not be able to do this. I think we want the factory to figure out from the url (maybe by even creating a client and attempting a ping) which adapter type to use.

clockard: Or we could update the sites object to have some sort of type that the user selects during creation.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
```diff
@@ -50,7 +50,7 @@
   private static final long STARTUP_DELAY = TimeUnit.MINUTES.toSeconds(1);
 
   private static final String DEFAULT_REPLICATION_PERIOD_STR =
-      String.valueOf(TimeUnit.MINUTES.toSeconds(5));
+      String.valueOf(TimeUnit.MINUTES.toSeconds(1));
```
clockard: Was this just for testing?

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapterFactory.java
peterhuffer: No it's not a guarantee, it's able to be customized. We make these assumptions for the REST and CSW endpoints.

replication-api/src/main/java/org/codice/ditto/replication/api/NodeAdapter.java
peterhuffer: I chose the simplest way to accomplish what we do already. Since we did not use the error info returned from catalog responses, I went with returning a boolean for now.

replication-api-impl/pom.xml
peterhuffer: Temporary.

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
```diff
@@ -0,0 +1,569 @@
...
+    metacards.forEach(
+        metacard ->
+            metacard.setAttribute(
+                new AttributeImpl(
+                    Core.RESOURCE_URI, ContentItem.CONTENT_SCHEME + ":" + metacard.getId())));
+
+    ddf.catalog.operation.UpdateRequest ddfUpdate = new UpdateRequestImpl(ids, metacards);
+
+    // todo: This previous state metacards is not correct, but it was this way before. Does this
+    // have any effect?
```
kcover: The CswStore doesn't do anything with these metacards it just hands them back to us in the response but the request will fail if you don't include them. 🙁 So we add this property just to keep the store happy.

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
```diff
@@ -0,0 +1,569 @@
...
+
+  @Override
+  public boolean deleteRequest(DeleteRequest deleteRequest) {
+    List<Metacard> metacards = getMetacards(deleteRequest.getMetadata());
+
+    DeleteRequestImpl ddfDeleteRequest =
+        new DeleteRequestImpl(
+            deleteRequest.getMetadata().stream().map(Metadata::getId).toArray(String[]::new));
+
+    // todo: this was not correct before either. we place versioned metacards in here.
```
kcover: Same as with the updates.

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
```diff
@@ -0,0 +1,569 @@
...
+import org.geotools.filter.text.cql2.CQLException;
+import org.geotools.filter.text.ecql.ECQL;
+import org.opengis.filter.Filter;
+import org.opengis.filter.sort.SortOrder;
+import org.osgi.framework.BundleContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** A {@link NodeAdapter} for interacting with DDF based systems. */
+public class DdfNodeAdapter extends AbstractCswStore implements NodeAdapter {
```
kcover: ✏️ Is this an adapter, or just a store for interacting with ddf?

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapterFactory.java
paouelle: 🤔

replication-api-impl/pom.xml
paouelle: as long as the final PR that gets merged takes that out and provides proper testing

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
```diff
@@ -0,0 +1,569 @@
...
+                .forEach(metadata::addLineage);
+          }
+
+          Attribute tags = metacard.getAttribute(Core.METACARD_TAGS);
+          if (tags != null) {
+            tags.getValues()
+                .stream()
+                .filter(String.class::isInstance)
+                .map(String.class::cast)
+                .forEach(metadata::addLineage);
```
kcover: addTag* instead of addLineage

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
peterhuffer: It's an adapter. It's pretty much what the Hybrid store was before.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/Syncer.java
```diff
@@ -0,0 +1,342 @@
...
+            doCreate(metadata);
+          }
+        } catch (VirtualMachineError e) {
+          throw e;
+        } catch (Exception e) {
+          final boolean sourceAvailable = source.isAvailable();
+          final boolean destinationAvailable = destination.isAvailable();
+          if (!sourceAvailable || !destinationAvailable) {
+            LOGGER.debug(
+                "Lost connection to either source {} (available={}) or destination {} (available={}). Setting status to {}",
```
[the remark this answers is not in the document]
peterhuffer: That is handled in the ReplicatorImpl. If there is no connection initially, it is a Connect Unavailable, if there is no connection here, it is Connection Lost.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/Syncer.java
```diff
@@ -0,0 +1,342 @@
...
+        LOGGER.trace(
+            "Sending update from {} to {} for metadata {}",
+            sourceName,
+            destinationName,
+            metadataId);
+
+        updated =
+            destination.updateRequest(new UpdateRequestImpl(Collections.singletonList(metadata)));
+      } else {
+        LOGGER.debug(
```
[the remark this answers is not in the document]
peterhuffer: I think we are less likely to this portion since we no longer rely on the last run success (and do the -1000) to query the remote system.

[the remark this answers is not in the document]
peterhuffer: Yeah I think we should no longer be able to hit this now that the last metadata modified is what is used to query the remote system for latest changes.

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfRestClientFactory.java
peterhuffer: Handled here: https://github.com/connexta/replication/pull/81

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapterFactory.java
peterhuffer: Handled here: https://github.com/connexta/replication/pull/81

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/Syncer.java
peterhuffer: Added synchronization around access to this variable.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/Syncer.java
peterhuffer: I added a javadoc note. But this status object will not be accessed by multiple threads because it is a new one each time, which once the job finishes is compounded with any existing status objects in the database for the given config.

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
```diff
@@ -0,0 +1,569 @@
...
+      cachedSystemName = identityMetacard.getResults().get(0).getMetacard().getTitle();
+    } else {
+      throw new AdapterException(
+          "No registry metacard available on remote node. Could not retrieve remote system name");
+    }
+    return cachedSystemName;
+  }
+
+  @Override
+  public QueryResponse query(QueryRequest queryRequest) {
```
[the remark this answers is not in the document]
peterhuffer: I put it this way cause I was getting frustrated jumping back and forth between the methods to figure out what was going on 😓 , but I agree there could be something done here to make the overall flow clearer.

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicationItemManager.java
```diff
@@ -0,0 +1,93 @@
...
+
+import java.util.List;
+import java.util.Optional;
+import org.codice.ditto.replication.api.ReplicationItem;
+
+public interface ReplicationItemManager {
+
+  /**
+   * If present, returns a {@link ReplicationItem} for the given {@link
+   * org.codice.ditto.replication.api.data.Metadata} identified by its.
```
kcover: ❓ is there a reason so many class links are fully qualified in this interfaces java docs?

peterhuffer: because they are not imported, and if they were imported the formatter would fail saying they are unused.

adapters/ddf/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
```diff
@@ -0,0 +1,595 @@
...
+    }
+  }
+
+  @Override
+  public boolean createRequest(CreateRequest createRequest) {
+    List<Metacard> metacards = getMetacards(createRequest.getMetadata());
+
+    try {
+      CreateResponse createResponse = super.create(new CreateRequestImpl(metacards));
+      return hasProcessingErrors(createResponse);
```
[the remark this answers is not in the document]
peterhuffer: yes! 🤦‍♂

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/CreateRequestImpl.java
```diff
@@ -0,0 +1,32 @@
...
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ditto.replication.api.impl.data;
+
+import java.util.List;
+import org.codice.ditto.replication.api.data.CreateRequest;
+import org.codice.ditto.replication.api.data.Metadata;
+
+public class CreateRequestImpl implements CreateRequest {
```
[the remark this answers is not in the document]
peterhuffer: I put only the response related classes in the common module since only the replication framework should be making requests between adapters.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/Syncer.java
```diff
@@ -0,0 +1,380 @@
...
+                "Lost connection to either source {} (available={}) or destination {} (available={}). Setting status to {}",
+                sourceName,
+                sourceAvailable,
+                destinationName,
+                destinationAvailable,
+                Status.CONNECTION_LOST);
+            replicationStatus.setStatus(Status.CONNECTION_LOST);
+            return new SyncResponse(replicationStatus.getStatus());
+          } else {
+            if (replicationItem.isPresent()) {
```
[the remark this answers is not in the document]
peterhuffer: nope

replication-api/src/main/java/org/codice/ditto/replication/api/NodeAdapterFactory.java
```diff
@@ -0,0 +1,10 @@
+package org.codice.ditto.replication.api;
+
+import java.net.URL;
+
+public interface NodeAdapterFactory {
```
paouelle: ✏️ javadoc

replication-api/src/main/java/org/codice/ditto/replication/api/NodeAdapterFactory.java
```diff
@@ -0,0 +1,10 @@
...
+
+  NodeAdapter create(URL url);
```
paouelle: ✏️ javadoc

replication-api/src/main/java/org/codice/ditto/replication/api/NodeAdapterFactory.java
```diff
@@ -0,0 +1,10 @@
...
+
+  NodeAdapterType getType();
```
paouelle: ✏️ javadoc

replication-api/src/main/java/org/codice/ditto/replication/api/persistence/ReplicationItemManager.java
```diff
@@ -0,0 +1,93 @@
...
+   * @param configId unique id for the {@link
+   *     org.codice.ditto.replication.api.data.ReplicatorConfig}
+   * @param startIndex index to start query at
+   * @param pageSize max number of results to return in a single query
+   * @return list of items for the given {@link
+   *     org.codice.ditto.replication.api.data.ReplicatorConfig} id
+   * @throws org.codice.ditto.replication.api.ReplicationPersistenceException if there is an error
+   *     fetching the items
+   */
+  List<ReplicationItem> getItemsForConfig(String configId, int startIndex, int pageSize);
```
paouelle: ✏️ Same comment as before about returning collections

clockard:
Hero passed ✅

PR #70, opened 2019-04-22
Fix missing metadata due to system time differences

peterhuffer:
#### What does this PR do?
- Records the last modified time of the last metacard to be processed, and uses that as the modified after date as part of the next query when determining the metacard change set to replicate.

@clockard @kcover @paouelle 

Setup a replication and do in order:

- Verify creates.
- Verify updates.
- Verify deletes.
- Verify create, update, and delete at once in the same run.

This was an issue when system time between machines varied. This solution makes it so the query does not rely on a different system time.

For example, if machine A is replicating to B, and machine A time is 1:30, but machine B time is 1:31, machine B will record its last success run at 1:31 and use that as its next query's modified after time. So machine A will receive a query, "give me metadata modified after 1:31", but if any new data is uploaded to A between 1:30 and 1:31, B will not receive those updates.

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
@@ -63,6 +63,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import javax.annotation.Nullable;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.shiro.SecurityUtils;
 import org.codice.ditto.replication.api.ReplicationException;
@@ -146,8 +147,9 @@ class SyncHelper {
       mcard = metacardResult.getMetacard();
       existingReplicationItem = persistentStore.getItem(mcard.getId(), sourceName, destinationName);
 
+      boolean deletedMetacard = isDeletedMetacard();
       try {
-        if (isDeletedMetacard()) {
+        if (deletedMetacard) {
           processDeletedMetacard();
         } else if (isUpdatable()) {
           processUpdate(existingReplicationItem.get());
@@ -163,6 +165,14 @@ class SyncHelper {
           recordItemFailure(e);
         }
       }
+
+      final Date modifiedDate;
+      if (deletedMetacard) {
+        modifiedDate = (Date) mcard.getAttribute(MetacardVersion.VERSIONED_ON).getValue();
+      } else {
+        modifiedDate = (Date) mcard.getAttribute(Core.METACARD_MODIFIED).getValue();
+      }
+      status.setLastMetadataModified(modifiedDate);
     }
     status.setStatus(canceled ? Status.CANCELED : Status.SUCCESS);
     return new SyncResponse(syncCount, failCount, bytesTransferred, status.getStatus());
@@ -199,13 +209,6 @@ class SyncHelper {
     final Filter ecqlFilter;
     final List<Filter> filters = createBasicMetacardFilters();
     final List<Filter> failureFilters = createFailedMetacardFilters();
-    final ReplicationStatus lastSuccessfulRun =
-        history
-            .getReplicationEvents(config.getName())
-            .stream()
-            .filter(s -> s.getStatus().equals(Status.SUCCESS))
-            .findFirst()
-            .orElse(null);
     Filter finalFilter;
 
     try {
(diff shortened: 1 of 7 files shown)


replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
```diff
@@ -236,6 +228,32 @@ private Filter buildFilter() {
     return finalFilter;
   }
 
+  @Nullable
+  private Date getModifiedAfter() {
+    final ReplicationStatus replicationStatus =
+        history
+            .getReplicationEvents(config.getName())
+            .stream()
+            .filter(s -> s.getStatus().equals(Status.SUCCESS))
```
paouelle:
```suggestion
            .map(ReplicationStatus::getStatus)
            .filter(Status.SUCCESS::equals)
```

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationStatusImpl.java
```diff
@@ -97,6 +99,17 @@ public void setLastSuccess(@Nullable Date lastSuccess) {
...
+  public void setLastMetadataModified(Date lastMetadataModified) {
+    this.lastMetadataModified = lastMetadataModified;
+  }
+
+  @Nullable
+  @Override
+  public Date getLastMetadataModified() {
+    return lastMetadataModified;
+  }
+
```
paouelle: ❓ Should we update the `toString() method?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
```diff
@@ -236,6 +228,32 @@ private Filter buildFilter() {
...
+            .findFirst()
+            .orElse(null);
+
+    if (replicationStatus == null) {
+      return null;
+    }
+
+    if (replicationStatus.getLastMetadataModified() != null) {
```
paouelle: ✏️ How about moving this logic inside the `ReplicationStatus` class and let it report the right date?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
peterhuffer: Can't map because I need the ReplicationStatus object after.

clockard: do we need the success check? if we are part way through a set and we lose connection it will not be successful but we would still have a valid last modified right? Could we instead just filter until we get the first status that has a last metadata modified set?

peterhuffer: True. I'll remove the check.

kcover:
Successful Hero ✔️

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
paouelle: ❓ Is it guaranteed that getReplicationEvents() returns the events on a given order? This code relies on that order to get the first. If that is the case then let's make sure we document in the `getReplicationEvents()` that they must be returned in that specific order. If that is not the case, then should we not sort them accordingly here?

peterhuffer: Since we condense the results of configuration runs, we should only have at most one status object per configuration.

paouelle: If that is the case then the why do we have a method that may return multiple? Since by design, it shouldn't happen, shouldn't we change that method to simply return one?

peterhuffer: We could, it's gone through some iteration. We used to have multiple status objects per configuration, but switched over to just a single condensed status. I think we should talk about how we are going to do status' of configurations though. I personally think multiple entries would be more valuable to an admin.

paouelle: Agreed. In addition, it will be much simpler and optimal in a cloud environment to have multiple entries and let the UI aggregate.

peterhuffer:
The master port for the changes in this PR will be included in https://github.com/connexta/replication/pull/67, which they were originally taken from.

PR #78, opened 2019-04-23
REP-77 Add root context to sites

### LINKED ISSUE #77: Add option to sites for specifying the root context path
**Describe the task**
Add option to sites for specifying the root context path. This will allow replication to work with sites that have changed their root context to something other than the default context of 'services'

clockard:
#### What does this PR do?
 Add root context to sites so replication works with systems with non default root contexts
Install and create a new site.
- verify that there is a root context field that is defaulted to 'services'
- verify that after creating the site the url shows the context
- verify that replication still works

Fixes #77 

Screen Shot 2019-04-23 at 2 55 10 PM

### DIFF
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.function.Consumer;
 import javax.annotation.Nullable;
 import javax.ws.rs.NotFoundException;
+import org.apache.commons.lang3.StringUtils;
 import org.codice.ddf.admin.api.fields.ListField;
 import org.codice.ddf.admin.common.fields.common.AddressField;
 import org.codice.ditto.replication.admin.query.replications.fields.ReplicationField;
@@ -43,6 +44,8 @@ public class ReplicationUtils {
 
   private static final long BYTES_PER_MB = 1024L * 1024L;
 
+  private static final String DEFAULT_CONTEXT = "services";
+
   private final SiteManager siteManager;
 
   private final ReplicatorConfigManager configManager;
@@ -63,8 +66,9 @@ public class ReplicationUtils {
   }
 
   // mutator methods
-  public ReplicationSiteField createSite(String name, AddressField address) {
-    String urlString = addressFieldToUrlString(address);
+  public ReplicationSiteField createSite(
+      String name, AddressField address, @Nullable String rootContext) {
+    String urlString = addressFieldToUrlString(address, rootContext);
     ReplicationSite newSite = siteManager.createSite(name, urlString);
     siteManager.save(newSite);
 
@@ -79,8 +83,9 @@ public class ReplicationUtils {
     return siteManager.exists(id);
   }
 
-  public ReplicationSiteField updateSite(String id, String name, AddressField address) {
-    String urlString = addressFieldToUrlString(address);
+  public ReplicationSiteField updateSite(
+      String id, String name, AddressField address, @Nullable String rootContext) {
+    String urlString = addressFieldToUrlString(address, rootContext);
     ReplicationSite site = siteManager.get(id);
 
     setIfPresent(site::setName, name);
@@ -89,10 +94,14 @@ public class ReplicationUtils {
     return getSiteFieldForSite(site);
   }
 
-  private String addressFieldToUrlString(AddressField address) {
+  private String addressFieldToUrlString(AddressField address, @Nullable String rootContext) {
+    final String context =
+        (rootContext != null) ? rootContext.trim().replaceFirst("^[/\\s]*", "") : DEFAULT_CONTEXT;
+
     return address.host().hostname() == null
         ? address.url()
-        : String.format("https://%s:%s", address.host().hostname(), address.host().port());
+        : String.format(
+            "https://%s:%s/%s", address.host().hostname(), address.host().port(), context);
   }
 
   public boolean deleteSite(String id) {
@@ -212,6 +221,7 @@ public class ReplicationUtils {
     address.hostname(url.getHost());
     address.port(url.getPort());
     field.address(address);
+    field.rootContext(StringUtils.isEmpty(url.getPath()) ? DEFAULT_CONTEXT : url.getPath());
     return field;
   }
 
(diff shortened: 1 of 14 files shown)


replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/persistence/SiteManagerImplTest.java
```diff
@@ -48,6 +48,7 @@
   @Before
   public void setUp() {
     System.setProperty("org.codice.ddf.system.siteName", "testSite");
+    System.setProperty("org.codice.ddf.system.rootContext", "services");
```
paouelle: ❗️ System properties should be cleaned up after the test has run. Use the RestoreSystemProperties JUnit rule from codice-test

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -89,10 +92,15 @@ public ReplicationSiteField updateSite(String id, String name, AddressField addr
     return getSiteFieldForSite(site);
   }
 
-  private String addressFieldToUrlString(AddressField address) {
+  private String addressFieldToUrlString(AddressField address, @Nullable String rootContext) {
+    String context = rootContext;
+    if (context == null) {
+      context = "services";
+    }
```
paouelle: ❗️ ❓ it looks like we are ignoring the context in some cases. Won't that be confusing?

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -19,6 +19,7 @@
 import java.util.function.Consumer;
 import javax.annotation.Nullable;
 import javax.ws.rs.NotFoundException;
+import org.apache.commons.lang.StringUtils;
```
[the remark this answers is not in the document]
clockard: Updated to use lang3

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -79,8 +81,9 @@ public boolean siteIdExists(String id) {
     return siteManager.exists(id);
   }
 
-  public ReplicationSiteField updateSite(String id, String name, AddressField address) {
-    String urlString = addressFieldToUrlString(address);
+  public ReplicationSiteField updateSite(
+      String id, String name, AddressField address, String rootContext) {
```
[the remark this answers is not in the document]
clockard: Updated

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -212,6 +220,7 @@ private ReplicationField getReplicationFieldForConfig(ReplicatorConfig config) {
     address.hostname(url.getHost());
     address.port(url.getPort());
     field.address(address);
+    field.rootContext(StringUtils.isEmpty(url.getPath()) ? "services" : url.getPath());
```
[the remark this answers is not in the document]
clockard: sure. changed

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/persistence/SiteManagerImplTest.java
clockard: added restore annotation

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/persist/CreateReplicationSite.java
```diff
@@ -38,6 +38,8 @@
```
peterhuffer: ❗️ We should update the DESCRIPTION of this method to be `Creates a replication site. If no rootContext is provided, it will default to "services"`.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
clockard: I guess the context will only be used if host/port are passed in to construct the url. If a url is passed in then we will just use it as is. I don't have a good explanation as to why we support both ways but our UI only uses the host/port/context option.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -89,10 +92,15 @@ public ReplicationSiteField updateSite(String id, String name, AddressField addr
     return getSiteFieldForSite(site);
   }
 
-  private String addressFieldToUrlString(AddressField address) {
+  private String addressFieldToUrlString(AddressField address, @Nullable String rootContext) {
+    String context = rootContext;
```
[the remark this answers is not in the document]
clockard: Updated

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/persist/CreateReplicationSite.java
clockard: Changed description

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -89,10 +94,14 @@ public ReplicationSiteField updateSite(String id, String name, AddressField addr
     return getSiteFieldForSite(site);
   }
 
-  private String addressFieldToUrlString(AddressField address) {
+  private String addressFieldToUrlString(AddressField address, @Nullable String rootContext) {
+    final String context =
+        (rootContext != null) ? rootContext.trim().replaceAll("^/*", "") : DEFAULT_CONTEXT;
```
peterhuffer: ❓ Should this be replace all, or just the first one if it is the first character in the context path?

clockard: Good point. This will do the right thing because of the "^" but it doesn't need the replace all.

paouelle: should we trim after the replace? or add use ( "^[/\\s]*")

paouelle: should we trim after the replace? or add use ( "^[/\s]*")

clockard: Updated to do a trim first and then run the suggested regex

peterhuffer:
Successful hero ✔️

PR #79, opened 2019-04-23
REP-75 Add modified time to configs

### LINKED ISSUE #75: Add a modified time to configurations
**Describe the task**
Add a field to the configuration class to hold the time that the configuration was last modified.

kcover:
#### What does this PR do?
Adds a modified field to the ReplicationConfig class which tells the last time the config was modified.

Install replication on a DDF where you can access solr.
Create a config and verify it has a modified date in solr.
Suspend the config and verify the modified date has changed.

Fixes #75 

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
@@ -126,6 +126,7 @@ public class ReplicationUtils {
     config.setFilter(filter);
     config.setBidirectional(biDirectional);
     config.setFailureRetryCount(5);
+    config.setModified();
     configManager.save(config);
 
     return getReplicationFieldForConfig(config);
@@ -145,6 +146,7 @@ public class ReplicationUtils {
     setIfPresent(config::setDestination, destinationId);
     setIfPresent(config::setFilter, filter);
     setIfPresent(config::setBidirectional, biDirectional);
+    config.setModified();
 
     configManager.save(config);
     return getReplicationFieldForConfig(config);
@@ -254,6 +256,7 @@ public class ReplicationUtils {
     }
 
     config.setSuspended(suspended);
+    config.setModified();
     configManager.save(config);
     if (suspended) {
       replicator.cancelSyncRequest(id);
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
@@ -13,6 +13,7 @@
  */
 package org.codice.ditto.replication.api.impl.data;
 
+import java.util.Date;
 import java.util.Map;
 import org.codice.ditto.replication.api.data.ReplicatorConfig;
 
@@ -41,9 +42,19 @@ public class ReplicatorConfigImpl extends AbstractPersistable implements Replica
 
   private static final String SUSPENDED_KEY = "suspended";
 
+  private static final String MODIFIED_KEY = "modified";
+
   /**
-   * 0/No Version - initial version of configs which were saved in the catalog framework. 1 - the
-   * first version of configs to be saved in the replication persistent store.
+   * Field specifying the version of the configuration. Possible versions include:
+   *
+   * <ol>
+   *   <li>0 (No version) - initial version of configs which were saved in the catalog framework
+   *   <li>1 - The first version of configs to be saved in the replication persistent store
+   *       <ul>
+   *         <li>Add <b>suspended</b> field of type boolean with default of false
+   *       </ul>
+   *   <li>2 - Add <b>modified</b> field of type Date
+   * </ol>
    */
   public static final int CURRENT_VERSION = 1;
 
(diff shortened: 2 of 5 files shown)


peterhuffer:
We should be adding the modified time to all configuration objects right? Sites, replications, replication items, etc.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -145,6 +146,7 @@ public ReplicationField updateReplication(
     setIfPresent(config::setDestination, destinationId);
     setIfPresent(config::setFilter, filter);
     setIfPresent(config::setBidirectional, biDirectional);
+    config.setModified();
```
paouelle: :question: Why not automatically set the modified timestamp inside the `save()` of all config objects?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -63,6 +74,8 @@
 
   private boolean suspended;
 
+  private Date modified;
```
paouelle: ✏️ I would suggest moving this attribute as part of the base class for all persistable

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -41,9 +42,19 @@
 
   private static final String SUSPENDED_KEY = "suspended";
 
+  private static final String MODIFIED_KEY = "modified";
+
   /**
-   * 0/No Version - initial version of configs which were saved in the catalog framework. 1 - the
-   * first version of configs to be saved in the replication persistent store.
+   * Field specifying the version of the configuration. Possible versions include:
```
peterhuffer: ❓ We can get rid of this section right? I think since we will be resetting back to version 1 in our new deployment, we won't have to worry about compatibility concerns with 0.2.x?

paouelle: This change needs to go in 0.2.x and I am not sure we will be able to ignore versioning from 0.2.x since in the future, we will be communicating with old replication-based nodes too. BTW, do we expose the version as part of the GraphQL API? If not, will that cause us problems when we try to communicate with a node or are we going to have to hardcode that a given old-replication-based version used a given set of versions for its objects? That might become messy when it is upgraded. 🤔

peterhuffer: No we don't expose it through GraphQL right now.

paouelle: We might want to consider exposing it

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
kcover: I assume that replication running on Ion would want to be able to save the config that it receives with the modified time that is already on the config. If I just timestamp it in the `save()` that timestamp will be overwritten.

paouelle: really good point. However, for Ion to be able to do that, the graphql endpoint would need to expose the ability to update the timestamp externally. Worst case, the timestamp would be updated automatically, Ion would reload the config, see it has changed and simply update Ion's DB; not a big issue

kcover: Aren't we updating the DB through the ReplicationPersistentStore? if we set the modified date in the `save()` method there, then we will have to find some other way to save the configs with the modified date we have.

paouelle: I am saying it might not matter that if Ion sends an update to config and that timestamp is updated to the latest which is newer than the remote. Actually, I might prefer that as I won't have to worry about time shifts between nodes.

peterhuffer: If I'm reading this right, it seems odd that we would allow the modified time to be updated via the graphql endpoint even though no other fields in the configuration changed.

kcover: That's a good point. I suppose somebody could make an update request with all null values, then we'd change the modified date for no good reason

paouelle: it goes beyond that, we would also be saving it back to the DB when in fact we shouldn't. I agree with @peterhuffer but the issue is bigger. We shouldn't be saving to the DB something that hasn;'t changed and if we don't then the modified timestamp wouldn't be updated which is why it shouldn't be under control of the client but under control of the save() operation that should be able to tell if anything has changed.

kcover:
Closing this because we want this change to apply to 0.2.x. We may port it to master later.

PR #82, opened 2019-04-26
REP-75 Add modified time to configs and sites

### LINKED ISSUE #75: Add a modified time to configurations

kcover:
#### What does this PR do?
Adds a modified time field to configs and sites and allows it to be read
through the graphql endpoint. Also makes the version of configs and sites readable from the graphql endpoint.

@peterhuffer @clockard @paouelle 

Install replication.
create sites and configurations.
query the graphql endoint for sites and configurations.
verify they contain accurate modified dates and versions.

Fixes #75 

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/fields/ReplicationField.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/fields/ReplicationField.java
@@ -58,6 +58,10 @@ public class ReplicationField extends BaseObjectField {
 
   private DateField lastSuccess;
 
+  private DateField modified;
+
+  private IntegerField version;
+
   public ReplicationField() {
     super("replication", "ReplicationConfig", DESCRIPTION);
     id = new PidField("id");
@@ -73,6 +77,8 @@ public class ReplicationField extends BaseObjectField {
     lastRun = new DateField("lastRun");
     firstRun = new DateField("firstRun");
     lastSuccess = new DateField("lastSuccess");
+    modified = new DateField("modified");
+    version = new IntegerField("version");
   }
 
   // setters
@@ -156,6 +162,21 @@ public class ReplicationField extends BaseObjectField {
     return this;
   }
 
+  public ReplicationField modified(String modified) {
+    this.modified.setValue(modified);
+    return this;
+  }
+
+  public ReplicationField modified(Date modified) {
+    this.modified.setValue(getInstantOrNull(modified));
+    return this;
+  }
+
+  public ReplicationField version(int version) {
+    this.version.setValue(version);
+    return this;
+  }
+
   private Instant getInstantOrNull(Date date) {
     if (date != null) {
       return date.toInstant();
@@ -236,6 +257,14 @@ public class ReplicationField extends BaseObjectField {
     return firstRun;
   }
 
+  public DateField modified() {
+    return modified;
+  }
+
+  public IntegerField version() {
+    return version;
+  }
+
   @Override
   public List<Field> getFields() {
     return ImmutableList.of(
@@ -251,7 +280,9 @@ public class ReplicationField extends BaseObjectField {
         suspended,
         lastRun,
         lastSuccess,
-        firstRun);
+        firstRun,
+        modified,
+        version);
   }
 
   public static class ListImpl extends BaseListField<ReplicationField> {
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/fields/ReplicationSiteField.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/fields/ReplicationSiteField.java
@@ -14,14 +14,18 @@
 package org.codice.ditto.replication.admin.query.sites.fields;
 
 import com.google.common.collect.ImmutableList;
+import java.time.Instant;
+import java.util.Date;
 import java.util.List;
 import java.util.concurrent.Callable;
 import org.codice.ddf.admin.api.Field;
 import org.codice.ddf.admin.common.fields.base.BaseListField;
 import org.codice.ddf.admin.common.fields.base.BaseObjectField;
+import org.codice.ddf.admin.common.fields.base.scalar.IntegerField;
 import org.codice.ddf.admin.common.fields.base.scalar.StringField;
 import org.codice.ddf.admin.common.fields.common.AddressField;
 import org.codice.ddf.admin.common.fields.common.PidField;
+import org.codice.ditto.replication.admin.query.replications.fields.DateField;
 
 public class ReplicationSiteField extends BaseObjectField {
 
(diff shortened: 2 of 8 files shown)


replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStore.java
```diff
@@ -82,12 +82,20 @@ public ReplicationPersistentStore(PersistentStore persistentStore) {
             String.format("Found two objects in persistence of type %s with ID %s", typeName, id));
       }
 
-      T persistable = createPersistable(klass);
-      persistable.fromMap(PersistentItem.stripSuffixes(storedMaps.get(0)));
-      return persistable;
+      T result = getPersistableFromMap(klass, storedMaps.get(0));
+      addModifiedTimeIfMissing(result);
```
kcover: Is it more readable if I do `result = addModifiedTimeIfMissing(result);`? or better the way it is?

admin/query/src/main/java/org/codice/ditto/replication/admin/query/replications/fields/ReplicationField.java
```diff
@@ -73,6 +75,7 @@ public ReplicationField() {
     lastRun = new DateField("lastRun");
     firstRun = new DateField("firstRun");
     lastSuccess = new DateField("lastSuccess");
+    modified = new DateField("modified");
```
paouelle: ❓ Can we also add the version of the object?

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/fields/ReplicationSiteField.java
```diff
@@ -71,6 +77,18 @@ public ReplicationSiteField rootContext(String context) {
     return this;
   }
 
+  public ReplicationSiteField modified(Date modified) {
+    this.modified.setValue(getInstantOrNull(modified));
+    return this;
+  }
+
+  private Instant getInstantOrNull(Date date) {
```
paouelle: ✏️ Should probably be static

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -55,6 +60,16 @@ public void setVersion(int version) {
     this.version = version;
   }
 
+  @Override
+  public Date getModified() {
+    return modified;
+  }
+
+  @Override
+  public void setModified() {
```
paouelle: ❓ Should we expose this one? I am thinking it should be removed or if left there for testing, then let's make it package private and annotate accordingly.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -65,6 +80,7 @@ public void setVersion(int version) {
     Map<String, Object> result = new HashMap<>();
     result.put(ID_KEY, getId());
     result.put(VERSION_KEY, getVersion());
+    result.put(MODIFIED_KEY, getModified());
```
paouelle: ❓ Should it be here that we update the modified timestamp?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -78,5 +94,6 @@ public void setVersion(int version) {
   public void fromMap(Map<String, Object> properties) {
     setId((String) properties.get(ID_KEY));
     setVersion((int) properties.get(VERSION_KEY));
+    this.modified = (Date) properties.get(MODIFIED_KEY);
```
paouelle: ✏️ For backward compatibility, I would check if the map doesn't contain the modified timestamp and initialize it for the first time here such that it never be `null`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStore.java
paouelle: ✏️ I would have moved that logic to the `fromMap()` above. It is better, in my opinion, to handle all version incompatibility stuff inside to/from map methods and let the rest work with an object upgraded at the latest version

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStore.java
```diff
@@ -82,12 +82,20 @@ public ReplicationPersistentStore(PersistentStore persistentStore) {
...
+      return result;
     }
   }
 
+  private <T extends AbstractPersistable> T addModifiedTimeIfMissing(T result) {
+    if (result.getModified() == null) {
+      result.setModified();
+      save(result);
```
paouelle: ✏️ I would not r-save it here. Leave it up to when it needs to really be modified.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStore.java
```diff
@@ -98,7 +106,10 @@ public ReplicationPersistentStore(PersistentStore persistentStore) {
           "An error occurred while retrieving data of type " + typeName, e);
     }
 
-    return storedMaps.stream().map(map -> getPersistableFromMap(klass, map));
+    return storedMaps
```
paouelle: ✏️ no need for this change if we let the `fromMap()` take care of incompatibilities in versions.

replication-api/src/main/java/org/codice/ditto/replication/api/data/Persistable.java
```diff
@@ -36,4 +38,14 @@
...
+
+  /**
+   * Returns a {@link Date} at which the persistable was last modified
+   *
+   * @return when the persistable was last modified
+   */
+  Date getModified();
+
+  /** Sets the modified date of this persistable to the current time */
+  void setModified();
```
paouelle: ✏️ I don't think we should expose this method. Instead, I would make sure we document the `save()` to indicate it will automatically update that field.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStore.java
```diff
@@ -119,6 +130,7 @@ String getPersistenceType(Class<? extends AbstractPersistable> klass) {
   }
 
   public void save(AbstractPersistable persistable) {
+    persistable.setModified();
```
paouelle: ✏️ Could be moved into the `toMap()` method such that handling of the modified timestamp be done internally inside the persistable item

peterhuffer:
Successful hero ✅ Verified accurate time stamps on new configurations/sites and that existing sites had modified times of `1970-01-01T00:00:00Z`

PR #88, opened 2019-05-01
Support disabling local replication by site

peterhuffer:
#### What does this PR do?
- Adds support for disabling local replication by site
- Addresses some tech-debt (unit tests, improvements to other graphql functions)

@clockard @paouelle @kcover 

Unit tests.

- Through a GraphiQL client, individually test updating a sites name, hostname, port and rootContext.

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
@@ -70,7 +70,7 @@ public class ReplicationUtils {
       String name, AddressField address, @Nullable String rootContext, boolean isRemoteManaged) {
     String urlString = addressFieldToUrlString(address, rootContext);
     ReplicationSite newSite = siteManager.createSite(name, urlString);
-    newSite.setIsRemoteManaged(isRemoteManaged);
+    newSite.setRemoteManaged(isRemoteManaged);
     siteManager.save(newSite);
 
     return getSiteFieldForSite(newSite);
@@ -89,10 +89,10 @@ public class ReplicationUtils {
     }
   }
 
-  public boolean isNotUpdatedSiteName(String id, String name) {
+  public boolean isUpdatedSitesName(String id, String name) {
     final ReplicationSite site = this.getSite(id);
     if (site != null) {
-      return !site.getName().equalsIgnoreCase(name);
+      return site.getName().equalsIgnoreCase(name);
     }
 
     return false;
@@ -103,26 +103,75 @@ public class ReplicationUtils {
   }
 
   public boolean updateSite(
-      String id, String name, AddressField address, @Nullable String rootContext) {
-    String urlString = addressFieldToUrlString(address, rootContext);
+      String id, @Nullable String name, AddressField address, @Nullable String rootContext) {
     ReplicationSite site = siteManager.get(id);
 
+    String updatedUrl = updateUrl(site.getUrl(), address, rootContext);
     setIfPresent(site::setName, name);
-    setIfPresent(site::setUrl, urlString);
+    setIfPresent(site::setUrl, updatedUrl);
 
     try {
       siteManager.save(site);
     } catch (ReplicationPersistenceException e) {
-      LOGGER.debug("Unable to saved updated site {} with id {}", site, id, e);
+      LOGGER.debug("Unable to save updated site {} with id {}", site, id, e);
       return false;
     }
 
     return true;
   }
 
+  private String updateUrl(String prevUrl, AddressField addressField, @Nullable String context) {
+    URL url;
+    try {
+      url = new URL(prevUrl);
+    } catch (MalformedURLException e) {
+      // sites previous saved url should always be valid
+      return "";
+    }
+
+    final String oldHostname = url.getHost();
+    final int oldPort = url.getPort();
+    final String oldContext = stipStartingSlashes(url.getPath());
+
+    final String newHostname;
+    final int newPort;
+    final String newContext;
+
+    if (addressField.host().hostname() != null) {
+      newHostname = addressField.host().hostname();
+    } else {
+      newHostname = oldHostname;
+    }
+
+    if (addressField.host().port() != null) {
+      newPort = addressField.host().port();
+    } else {
+      newPort = oldPort;
+    }
+
+    if (context != null) {
+      newContext = stipStartingSlashes(context);
+    } else {
+      newContext = oldContext;
+    }
+
+    try {
+      return new URL(
+              String.format("%s://%s:%d/%s", url.getProtocol(), newHostname, newPort, newContext))
+          .toString();
+    } catch (MalformedURLException e) {
+      // all fields are validated by graphql before we make it here so this cannot be hit.
+      return "";
+    }
+  }
+
+  private String stipStartingSlashes(String str) {
+    return str.trim().replaceFirst("^[/\\s]*", "");
+  }
+
   private String addressFieldToUrlString(AddressField address, @Nullable String rootContext) {
     final String context =
-        (rootContext != null) ? rootContext.trim().replaceFirst("^[/\\s]*", "") : DEFAULT_CONTEXT;
+        (rootContext != null) ? stipStartingSlashes(rootContext) : DEFAULT_CONTEXT;
 
     return address.host().hostname() == null
         ? address.url()
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
@@ -67,10 +67,10 @@ public class ReplicationUtils {
 
   // mutator methods
   public ReplicationSiteField createSite(
-      String name, AddressField address, @Nullable String rootContext, boolean isDisableLocal) {
+      String name, AddressField address, @Nullable String rootContext, boolean isRemoteManaged) {
     String urlString = addressFieldToUrlString(address, rootContext);
     ReplicationSite newSite = siteManager.createSite(name, urlString);
-    newSite.setIsDisabledLocal(isDisableLocal);
+    newSite.setIsRemoteManaged(isRemoteManaged);
     siteManager.save(newSite);
 
     return getSiteFieldForSite(newSite);
@@ -80,6 +80,7 @@ public class ReplicationUtils {
     return siteManager.objects().map(ReplicationSite::getName).anyMatch(name::equalsIgnoreCase);
   }
 
+  @Nullable
   public ReplicationSite getSite(String id) {
     try {
       return siteManager.get(id);
@@ -89,7 +90,7 @@ public class ReplicationUtils {
   }
 
   public boolean isNotUpdatedSiteName(String id, String name) {
-    ReplicationSite site = this.getSite(id);
+    final ReplicationSite site = this.getSite(id);
     if (site != null) {
       return !site.getName().equalsIgnoreCase(name);
     }
@@ -101,21 +102,22 @@ public class ReplicationUtils {
     return siteManager.exists(id);
   }
 
-  public ReplicationSiteField updateSite(
-      String id,
-      String name,
-      AddressField address,
-      @Nullable String rootContext,
-      boolean isDisabledLocal) {
+  public boolean updateSite(
+      String id, String name, AddressField address, @Nullable String rootContext) {
     String urlString = addressFieldToUrlString(address, rootContext);
     ReplicationSite site = siteManager.get(id);
 
     setIfPresent(site::setName, name);
     setIfPresent(site::setUrl, urlString);
-    site.setIsDisabledLocal(isDisabledLocal);
-    siteManager.save(site);
 
-    return getSiteFieldForSite(siteManager.get(site.getId()));
+    try {
+      siteManager.save(site);
+    } catch (ReplicationPersistenceException e) {
+      LOGGER.debug("Unable to saved updated site {} with id {}", site, id, e);
+      return false;
+    }
+
+    return true;
   }
 
   private String addressFieldToUrlString(AddressField address, @Nullable String rootContext) {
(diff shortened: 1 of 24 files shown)


peterhuffer:
There is a timing issue when updating a ReplicationSite and returning the updated object. This is where it has not finished being saved in solr, therefore when we fetch the updated one, we do not always get the latest modified timestamp or updated fields. I would like to change the UpdateReplicationSite to return a boolean if there are no objections. This would mean that a client would have to re-query all the sites, since we do not support querying sites by ID - so I'd probably add that support as well.

paouelle:
@peterhuffer This seems like a better design for me especially in cloud-based applications where data saved might also not be available right away because of eventual consistency.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -67,31 +67,55 @@ public ReplicationUtils(
...
+  public ReplicationSite getSite(String id) {
+    try {
+      return siteManager.get(id);
+    } catch (NotFoundException e) {
+      return null;
+    }
+  }
+
+  public boolean isNotUpdatedSiteName(String id, String name) {
+    ReplicationSite site = this.getSite(id);
```
paouelle: ✏️ [Nit] You could make it final and clarify for all that the reference will not changed in the method after.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/persist/CreateReplicationSite.java
```diff
@@ -43,21 +44,27 @@
...
     super(FIELD_NAME, DESCRIPTION);
 
     this.replicationUtils = replicationUtils;
-    name = new StringField("name");
-    address = new AddressField();
-    rootContext = new StringField("rootContext");
-    name.isRequired(true);
-    address.isRequired(true);
-    rootContext.isRequired(true);
+    this.name = new StringField("name");
```
paouelle: Thank you

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
```diff
@@ -29,9 +29,13 @@
 
   private static final String URL_KEY = "url";
 
+  private static final String IS_DISABLED_LOCAL_KEY = "is-disabled-local";
+
   /** 1 - initial version. */
   public static final int CURRENT_VERSION = 1;
```
paouelle: ❗️ I think we should bump the version here and default the flag to false when retrieving a version 1

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -101,21 +102,22 @@ public boolean siteIdExists(String id) {
...
     setIfPresent(site::setName, name);
     setIfPresent(site::setUrl, urlString);
-    site.setIsDisabledLocal(isDisabledLocal);
-    siteManager.save(site);
 
-    return getSiteFieldForSite(siteManager.get(site.getId()));
+    try {
+      siteManager.save(site);
+    } catch (ReplicationPersistenceException e) {
+      LOGGER.debug("Unable to saved updated site {} with id {}", site, id, e);
```
paouelle: ❓ Should we not log at error or warning level the issue and at debug the stack trace?

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/discover/GetReplicationSites.java
```diff
@@ -21,39 +21,54 @@
...
 import org.codice.ditto.replication.admin.query.sites.fields.ReplicationSiteField;
+import org.codice.ditto.replication.api.data.ReplicationSite;
 
 public class GetReplicationSites extends BaseFunctionField<ListField<ReplicationSiteField>> {
 
   public static final String FIELD_NAME = "sites";
 
   public static final String DESCRIPTION =
-      "Retrieves all the sites currently configured for replication.";
+      "Retrieves all the sites currently configured for replication, or a single site identified by a pid.";
```
paouelle: ✏️ Instead of `a single site ...`, maybe something like `a site given its identifier`.

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/persistence/ReplicationPersistentStoreTest.java
```diff
@@ -82,7 +82,7 @@
 
   private static final String DELETE_DATA = "deleteData";
 
-  private static final String IS_DISABLED_LOCAL = "is-disabled-local";
+  private static final String IS_DISABLED_LOCAL = "is-remote-managed";
```
paouelle: ✏️ `IS_REMOTE_MANAGED`

replication-api/src/main/java/org/codice/ditto/replication/api/data/ReplicationSite.java
```diff
@@ -45,12 +45,12 @@
...
   /**
-   * See {@link #isDisabledLocal()}.
+   * See {@link #isRemoteManaged()}.
    *
-   * @param isDisabledLocal whether or not the local process is responsible for running replications
+   * @param isRemoteManaged whether or not the local process is responsible for running replications
    *     this site is associated with.
    */
-  void setIsDisabledLocal(boolean isDisabledLocal);
+  void setIsRemoteManaged(boolean isRemoteManaged);
```
paouelle: ✏️ `setRemoteManaged()`

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/fields/ReplicationSiteField.java
```diff
@@ -47,7 +47,7 @@
 
   private IntegerField version;
 
-  private IsDisabledLocalField isDisabledLocal;
+  private IsRemoteManagedField isDisabledLocal;
```
paouelle: ✏️ remoteManaged` instead of `isDisabledLocal`

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/fields/IsRemoteManagedField.java
```diff
@@ -15,14 +15,14 @@
 
 import org.codice.ddf.admin.common.fields.base.scalar.BooleanField;
 
-public class IsDisabledLocalField extends BooleanField {
+public class IsRemoteManagedField extends BooleanField {
```
paouelle: ✏️ Should we just name the class `RemoteManageField`? `is` is a prefix for a getter method.
❗️ javadoc 😄

peterhuffer:
There are a few issues I have to fix when I was testing updating a site. Unrelated to changes in PR, but I will fix it as part of this PR.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/fields/IsRemoteManagedField.java
```diff
@@ -0,0 +1,28 @@
...
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package org.codice.ditto.replication.admin.query.sites.fields;
+
+import org.codice.ddf.admin.common.fields.base.scalar.BooleanField;
+
+public class IsRemoteManagedField extends BooleanField {
```
kcover: Why not just use a BooleanField?

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/persist/DeleteReplicationSite.java
```diff
@@ -60,8 +62,12 @@ public void validate() {
       return;
     }
 
-    ListField<ReplicationField> repFields = replicationUtils.getReplications(true);
     String idToDelete = id.getValue();
+    if (!replicationUtils.siteIdExists(idToDelete)) {
+      addErrorMessage(DefaultMessages.noExistingConfigError());
+    }
```
kcover: replicationUtils.deletesite() returns a boolean based on whether the id was found or not. You could also argue that if the site doesn't exist then it was probably already deleted and the user doesn't need to worry about it.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/fields/IsRemoteManagedField.java
peterhuffer: I did this so that a better description for this field could be provided in the graphql schema because boolean field does not have a ctor that takes a description unfortunately.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/persist/UpdateReplicationSite.java
```diff
@@ -71,12 +76,31 @@ public ReplicationSiteField getReturnType() {
...
+    }
+
+    final String updatedName = name.getValue();
+    final String updateId = id.getValue();
+    if (!replicationUtils.siteIdExists(updateId)) {
+      addErrorMessage(DefaultMessages.noExistingConfigError());
+    }
+    if (updatedName != null
+        && (replicationUtils.isNotUpdatedSiteName(updateId, updatedName)
+            || replicationUtils.isDuplicateSiteName(updatedName))) {
```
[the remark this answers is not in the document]
peterhuffer: If an update request is sent with the name of the site identified by the ID, the isDuplicateSiteName would not be sufficient because it would return true for its own name, but it should be allowed to update regardless if it is its own name. However this logic is wrong and I updated it :D

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
peterhuffer: I do not think so because at warn, the admin cannot do anything to fix it. And at error it's not really a critical error that stops the system.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/persist/DeleteReplicationSite.java
paouelle: I agree with @kcover here. If the site did not exist to start with then the end result is the same as if it did and therefore it shouldn't be an error but more of a no-op

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
paouelle: That is why I was suggesting warning and it is also why there are levels such as critical and error. Warning which typically indicates that something happened that won't prevent the system from continuing to function. Without at least logging it at warning, how would an admin know something happened? We shouldn't need to turn on debug to see stuff happened like this. We shou;d turn debug to get more info about stuff that happened.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
```diff
@@ -108,13 +121,32 @@ void scheduleReplication() {
...
+
+  private boolean sourceOrDestinationIsRemotelyManaged(ReplicatorConfig replicatorConfig) {
+    try {
+      return siteManager.get(replicatorConfig.getSource()).isRemoteManaged()
+          || siteManager.get(replicatorConfig.getDestination()).isRemoteManaged();
+    } catch (NotFoundException | ReplicationPersistenceException e) {
+      LOGGER.debug(
+          "Unable to determine if replication {} should be run based on its sites. This replication will not be run.",
+          replicatorConfig.getName());
+      return true;
```
clockard: Return true here? Should we just catch the error in the calling method and log it?

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationUtils.java
```diff
@@ -67,36 +67,111 @@ public ReplicationUtils(
 
   // mutator methods
   public ReplicationSiteField createSite(
-      String name, AddressField address, @Nullable String rootContext) {
+      String name, AddressField address, @Nullable String rootContext, boolean isRemoteManaged) {
```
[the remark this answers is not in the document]
paouelle: I completely missed that one. Good catch.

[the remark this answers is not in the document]
peterhuffer: The graphql create site doesn't take in the argument, it just always passes in false. I thought we wanted it to only be updated internally though after doing a ping to the site to find out if the heartbeat is implemented?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
peterhuffer: Depends, do we still want it to run locally if we are not able to determine by the sites?

paouelle: I am not sure it's one or the other. When we finally make the decision to send a heartbeat it would be neither of these 2 since I don't think we want to start replication since we were unable to load the site config anyway and we are unable to send the heartbeat because ... we couldn't load the site anyway.
Does it really make sense to have this method to start with knowing what is coming? I think we might just want to get a hold of the 2 sites in the above method and make our decision there based on one or the other because down the road (next PR), we will need to save the site we need to send a heartbeat in a set such that after the loop we can send the heartbeats (or request them to be sent).

admin/query/src/main/java/org/codice/ditto/replication/admin/query/sites/persist/DeleteReplicationSite.java
peterhuffer: The UI can treat it as a noop, but I think from and API perspective it would be important to know if the operation actually did something.

clockard:
Heroed ✅

clockard:
heroed ✅

PR #90, opened 2019-05-09
[0.2.x] REP-87 Fix lastRun and lastSuccess not auto updating

### LINKED ISSUE #87: Last run and last success do not auto update in the UI
**Describe the bug**
The last run and last success dates do not properly update in the UI. This is because the actual date is received as UTC, and it is not changing, but the way we display it changes. React does not re-render components if the data does not change.

**Affects version**
0.2.2

**To Reproduce**
1. Create a replication, let it run, then disable it.
2. Observe the last run and last success do not update without refreshing the page.

**Expected behavior**
The last run and last success relative time displays to update automatically.

**Desktop (please complete the following information):**
 - Browser: chrome

**Additional context**
One solution is to store the relative time for those 2 dates in the local state of a table row and use a timer to update that state every minute.

peterhuffer:
#### What does this PR do?
Adds a 1 minute interval that will update the relative time display of replication configurations even if their lastRun and lastSuccess dates don't change on polling.

@clockard @kcover @paouelle 

- Setup 2 replications (to verify timers are independent)
- Make sure the date columns tick independently (can put one replication in inactive)

Fixes #87 

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/ui/src/main/webapp/components/replications/ReplicationsTable.js
+++ b/ui/src/main/webapp/components/replications/ReplicationsTable.js
@@ -30,6 +30,7 @@ import Immutable from 'immutable'
 import { withStyles } from '@material-ui/core/styles'
 import ActionsMenu from './ActionsMenu'
 import Replications from './replications'
+import ReactInterval from 'react-interval'
 
 const styles = {
   root: {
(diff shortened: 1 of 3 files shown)


ui/src/main/webapp/components/replications/ReplicationsTable.js
```diff
@@ -58,6 +61,53 @@ class ReplicationsTable extends React.Component {
     this.setState({ anchor: null })
   }
 
+  render() {
+    const { replication } = this.props
+
+    return (
+      <TableRow>
+        <ReactInterval
+          enabled={true}
```
peterhuffer: ❗️ Add interval property here.

kcover:
Successful Hero! ✔️

PR #99, opened 2019-06-03
[WIP] Support heartbeating and verifying nodes

peterhuffer:
#### What does this PR do?
Unit tests coming.

- Add back-end support for verifying sites
- Add front-end support for displaying remote managed sites and configurations

@paouelle @clockard @kcover 

See mee.

In order for Replication to work in the future with Ion, Ion nodes need to be identified.

Fixes

image
image

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/replication-api-impl/pom.xml
+++ b/replication-api-impl/pom.xml
@@ -25,7 +25,7 @@
     <packaging>bundle</packaging>
 
     <properties>
-        <jackson.version>2.9.8</jackson.version>
+        <jackson.version>2.9.9</jackson.version>
     </properties>
 
     <dependencies>
@@ -265,9 +265,29 @@
             <version>${jackson.version}</version>
         </dependency>
         <dependency>
-            <groupId>ddf.platform</groupId>
-            <artifactId>branding-api</artifactId>
-            <version>${ddf.version}</version>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.jaxrs</groupId>
+            <artifactId>jackson-jaxrs-base</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-jaxb-annotations</artifactId>
+            <version>${jackson.version}</version>
         </dependency>
         <!-- cxf discovery client -->
     </dependencies>
@@ -312,25 +332,15 @@
                         <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
                         <Import-Package>
                             !org.jvnet.staxex,
-                            !io.swagger.annotations,
                             ddf.security.permission,
                             ddf.security.encryption;version="[1.0.0, 3.0.0]",
+                            !io.swagger.annotations,
                             *
                         </Import-Package>
                         <Export-Package>
                             org.codice.ditto.replication.api.impl.data,
                             org.codice.ditto.replication.api.impl.mcard,
                             org.codice.ditto.replication.api.impl.modern,
-
-                            <!-- cxf discovery client -->
-                            com.connexta.discovery.rest.cxf.client,
-                            com.connexta.discovery.rest.models,
-                            com.fasterxml.jackson.databind.*,
-                            com.fasterxml.jackson.core.*,
-                            com.fasterxml.jackson.annotation,
-                            com.fasterxml.jackson.jaxrs.*,
-                            com.fasterxml.jackson.module.jaxb.*
-                            <!-- cxf discovery client -->
                         </Export-Package>
                         <Embed-Dependency>
                             spatial-csw-source-common,
@@ -352,7 +362,13 @@
                             <!-- cxf discovery client -->
                             discovery-api-cxf-client,
                             discovery-api-rest-models,
-                            jackson-datatype-jsr310 <!-- Allows jackson parsing of OffsetDateTime -->
+                            jackson-datatype-jsr310, <!-- Allows jackson parsing of OffsetDateTime -->
+                            jackson-annotations,
+                            jackson-core,
+                            jackson-databind,
+                            jackson-jaxrs-json-provider,
+                            jackson-jaxrs-base,
+                            jackson-module-jaxb-annotations
                             <!-- cxf discovery client -->
                         </Embed-Dependency>
                         <Embed-Transitive>true</Embed-Transitive>
--- a/replication-api-impl/pom.xml
+++ b/replication-api-impl/pom.xml
@@ -24,6 +24,10 @@
     <name>Replication :: API Impl</name>
     <packaging>bundle</packaging>
 
+    <properties>
+        <jackson.version>2.9.8</jackson.version>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>replication</groupId>
@@ -238,6 +242,34 @@
             <version>${ddf.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <!-- cxf discovery client -->
+        <dependency>
+            <groupId>com.connexta.discovery</groupId>
+            <artifactId>discovery-api-cxf-client</artifactId>
+            <version>${discovery-api.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.connexta.discovery</groupId>
+            <artifactId>discovery-api-rest-models</artifactId>
+            <version>${discovery-api.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.jaxrs</groupId>
+            <artifactId>jackson-jaxrs-json-provider</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>ddf.platform</groupId>
+            <artifactId>branding-api</artifactId>
+            <version>${ddf.version}</version>
+        </dependency>
+        <!-- cxf discovery client -->
     </dependencies>
     <build>
         <resources>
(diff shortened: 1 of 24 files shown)


replication-api-impl/pom.xml
```diff
@@ -24,6 +24,10 @@
     <name>Replication :: API Impl</name>
     <packaging>bundle</packaging>
 
+    <properties>
+        <jackson.version>2.9.8</jackson.version>
```
clockard: Version 2.9.9 is available

replication-api-impl/pom.xml
```diff
@@ -280,14 +312,25 @@
...
                             *
                         </Import-Package>
                         <Export-Package>
                             org.codice.ditto.replication.api.impl.data,
                             org.codice.ditto.replication.api.impl.mcard,
-                            org.codice.ditto.replication.api.impl.modern
+                            org.codice.ditto.replication.api.impl.modern,
+
+                            <!-- cxf discovery client -->
+                            com.connexta.discovery.rest.cxf.client,
```
clockard: I know there was some class path issues going on but this still doesn't seem like the right thing to do.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
```diff
@@ -73,23 +135,307 @@ public void cleanUp() {
...
+    ContactInfo contactInfo = new ContactInfo().email(SystemInfo.getSiteContatct());
+    com.connexta.discovery.rest.models.SystemInfo systemInfo =
+        new com.connexta.discovery.rest.models.SystemInfo()
+            .contact(contactInfo)
+            .id(System.getProperty("org.codice.ddf.system.registry-id"))
+            .name(SystemInfo.getSiteName())
+            .organization(SystemInfo.getOrganization())
+            .version(SystemInfo.getVersion())
+            .url(SystemBaseUrl.EXTERNAL.getBaseUrl())
+            .product(brandingPlugin.getProductName());
```
clockard: Why do we need the branding plugin for this? Can we just use the system property `org.codice.ddf.system.branding`? I would like to avoid pulling in more ddf dependencies if possible.

replication-api-impl/pom.xml
```diff
@@ -238,6 +242,34 @@
...
+            <groupId>com.fasterxml.jackson.jaxrs</groupId>
+            <artifactId>jackson-jaxrs-json-provider</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+            <version>${jackson.version}</version>
+        </dependency>
+        <dependency>
```
clockard: See comment later on about removing this dependency.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
```diff
@@ -73,23 +135,307 @@ public void cleanUp() {
...
 
-    // For now, fake the verification until implemented
-    site.setVerifiedUrl(url);
-    siteManager.save(site);
+      if (heartbeatResult.isHeartbeatSupported()) {
+        site.setRemoteManaged(true);
+        site.setVerifiedUrl(url);
+        siteManager.save(site);
+      } else if (heartbeatResult.isTmpRedirect()) {
+        HeartbeatResult redirectResult = doDiscovery(heartbeatResult.getRedirectUrl());
```
clockard: if the initial heartbeat call is successful and give us a new url do we need to re-run discovery at that point? Seems like we have verified it is an ION system at that point so we should just set the site properties and save it regardless of whether or not we can verify the redirect.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
```diff
@@ -73,23 +135,307 @@ public void cleanUp() {
...
+        if (redirectResult.isHeartbeatSupported()) {
+          site.setRemoteManaged(true);
+          site.setVerifiedUrl(url);
+          siteManager.save(site);
+        }
+        // do nothing and retry the heartbeat next time
+      } else if (heartbeatResult.isPermRedirect()) {
```
clockard: Why do we have a perm and tmp redirect. Do we care? Doesn't look like we will attempt to reverify either way.

paouelle:
We need to keep track in memory of sites that we sent a heartbeat to and we failed because of a bad request such that we do not retry until we restarted (possibly keep track of these sites in memory). For sites that responds with an unsupported API version or internal error, I think we can let the main loop retry these sites a minute later. For redirects (permanent or temporary), we should retry right away on the spot and follow all these redirects until we reach a maximum configurable number of follows. If it fails, let the loop retry 1 minute later. This is typically how http client handles following redirects.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
```diff
@@ -73,23 +135,307 @@ public void cleanUp() {
...
+              if (statusOptional.isPresent()) {
+                ReplicationStatus replicationStatus = statusOptional.get();
+                updater.accept(replicationStatus, replicationStatus.getStatus());
+
+                try {
+                  // todo this is going to undesirably condense the status' stats...
+                  history.addReplicationEvent(replicationStatus);
+                } catch (ReplicationPersistenceException e) {
+                  LOGGER.debug(
+                      "Failed to update status while heartbeating site {}", site.getUrl(), e);
```
paouelle: ❓ Should we logged this as `warning` level instead of debug?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
```diff
@@ -73,23 +135,307 @@ public void cleanUp() {
...
+                }
+              } else {
+                LOGGER.warn("No existing status to update for config {}", config.getName());
```
paouelle: ❓ I think we need to create one otherwise the first time around we are trying to verify a site will never be reported as CONNECTION_UNAVAILABLE.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
```diff
@@ -73,23 +135,307 @@ public void cleanUp() {
...
+        clientFactoryFactory.getSecureCxfClientFactory(
+            url,
+            DiscoveryApi.class,
+            Collections.singletonList(
+                new JacksonJaxbJsonProvider(
+                    objectMapper, JacksonJaxbJsonProvider.DEFAULT_ANNOTATIONS)),
+            null,
+            true /* disableCnCheck */,
+            false /* allowRedirects */,
+            30000 /* connectionTimeout */,
```
paouelle: ❓ Should these timeouts be constants or better yet system properties?
❓ Is 30 seconds not a long time to wait?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
```diff
@@ -140,6 +140,21 @@ public void fromMap(Map<String, Object> properties) {
     }
   }
 
+  @Override
+  public int hashCode() {
```
paouelle: ✏️ Maybe something that should be pushed to the base class?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
```diff
@@ -140,6 +140,21 @@ public void fromMap(Map<String, Object> properties) {
...
+    return getId().hashCode();
+  }
+
+  @Override
+  public boolean equals(Object obj) {
```
paouelle: ✏️ Maybe something that should be pushed to the base class?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
peterhuffer: I went through the branding plugin cause it was what I was familiar with. We can go to the system property, yes. There is a small difference though. By default the plugin gives something like `DDF 1.2.3` but that system property will give `ddf` by default. 

It doesn't matter if we add more DDF dependencies here though cause this PR is against 0.2.x and will not be ported to master.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
peterhuffer: If it is a permanent or temporary redirect, a heartbeat is sent again to the redirect URL. If we hit a proxy we will need to follow the redirect.

paouelle: Correct. All redirects must be followed by designed (HTTP spec) up to a max number of redirects. We care about permanent a bit more as we need to save the new location for subsequent messages.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
peterhuffer: 30 seconds is typically what we use (and is the default of the CxfClientFactory) due to limitations of the environment.
I'd prefer not to make more configuration for these, but if we feel we need to we can.

paouelle: Not knowing what will happen on certain networks, I would feel better if we had a way to have our customer tweaks these as opposed to delivering a code patch. We can read system properties while defaulting in code to 30 secs.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/HeartbeaterAndVerifierImpl.java
```diff
@@ -73,23 +161,404 @@ public void cleanUp() {
...
+    do {
+
+      boolean redirected = handleHeartbeatResult(heartbeatResult, site);
+      if (redirected) {
+        url = heartbeatResult.getRedirectUrl();
+        numRedirects++;
+
+        if (heartbeatResult.isPermRedirect()) {
+          site.setRemoteManaged(true);
+          site.setVerifiedUrl(url);
```
peterhuffer: This will set the verified URL as the perm redirect URL every time, but I just realized technically this is not a verified URL until we test the redirect URL so probably cannot save it every time.

paouelle: Because it is a permanent redirect we have to save it. What we cannot do is assumed it is remotely managed yet.

peterhuffer:
Closing this PR in favor of https://github.com/connexta/replication/pull/122

PR #118, opened 2019-06-13
Remove the docs in master

### LINKED ISSUE #114: Configure docs to be pulled into upstream projects
**Describe the task**
Update the documentation so that it can be pulled into upstream projects via the maven artifact.

kcover:
#### What does this PR do?
The docs are no longer relevant in master so they have been removed.

CI build

#### Checklist:
- [x] Documentation Updated 😂  
- [ ] Update / Add Unit Tests


docs/src/main/resources/content/replication-installing.adoc
```diff
@@ -1,16 +1,16 @@
...
-:order: 02
+:order: 05
 
 == {title}
 
 Replication is not installed by default with a standard installation. There are several installation options available.
 
 === Installation When Replication _*Is*_ Included In Distribution
-==== Installing Via Feature
+==== Installing Via Admin Console
```
clockard: Do we want to say Admin UI instead of Console?

docs/src/main/resources/content/_using/replication-limitations.adoc
```diff
@@ -1,33 +0,0 @@
-:title: Known Issues, Limitations, and Assumptions
```
peterhuffer: ❓ Does this limitations documentation still have value? These are technically still applicable with master. Should it become part of the README?

docs/src/main/resources/content/_using/replication.adoc
```diff
@@ -1,15 +0,0 @@
-:title: Introduction
-:type: using
-:status: published
-:summary: Instructions for performing replication.
-:order: 01
-
-== {title}
-
-Replication is the process of creating a copy of a subset of data and storing it on another
```
peterhuffer: ❓ @clockard Can we move this overview of Replication to the README?

clockard: Yes. I can add it to the readme

docs/src/main/resources/content/_using/replication-limitations.adoc
kcover: Everything except for derived resources is a limitation imposed by federation, but that shouldn't really be an issue anymore because they should just be querying and replicating from ION right? I'm not sure if derived resources is an issue either because the derived resources should be the same when generated on either side right?

peterhuffer: Ya in the future many of these will go away. In the meantime harvesting from existing DDF sites will still have these issues.

clockard: I included the limitations in the readme

PR #120, opened 2019-06-24
[0.2.x] REP-119 Fix failure retry query bug

### LINKED ISSUE #119: When failure retries are included in csw query no new items are returned
**Describe the bug**
No new items are returned when querying over csw if the query contains search parameters for previously failed items. The matching failed items are returned however. 

The expected behavior is that new items would also be returned when we include failure items in the query as well. 

**Affects version**
0.2.x, 0.3.0

**To Reproduce**
1. Attempt to replicate a product that will fail
2. Ingest a new item that matches the filter criteria
3. attempt to replicate again
4. notice that the new item has not been replicated as it should be

**Expected behavior**
products matching the given filter will be replicated even when there are items in the repliction that previously failed.

kcover:
#### What does this PR do?
This commit fixes a bug where no new items are returned when failure
retries are included in a csw query to DDF

-Install a system based on a released DDF 2.13.x
-Create a replication and set an item up to fail (the only way i know of to do this is to create a cdm     without configuring the resource reader for the same directory).
-Once you have a failed item, ingest a new product through the UI and replicate again.
-Verify the product ingested through the UI was replicated.

Fixes #119 

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
@@ -114,7 +114,7 @@ class SyncHelper {
 
   private volatile boolean canceled = false;
 
-  private ReplicationStatus status;
+  private final ReplicationStatus status;
 
   public SyncHelper(
       ReplicationStore source,
@@ -139,9 +139,26 @@ class SyncHelper {
   }
 
   public SyncResponse sync() {
-    for (Result metacardResult : getMetacardChangeSet()) {
+    // Failed items and new items are replicated separately due to a bug in DDF that was only fixed
+    // recently
+    List<Filter> failedItemFilters = createFailedMetacardFilters();
+    if (!failedItemFilters.isEmpty()) {
+      replicateSubset(builder.anyOf(failedItemFilters));
+    }
+
+    // if the connection was lost while replicating failed items don't continue replicating
+    if (!canceled && status.getStatus() != Status.CONNECTION_LOST) {
+      replicateSubset(buildChangeSetFilter());
+    }
+
+    status.setStatus(canceled ? Status.CANCELED : Status.SUCCESS);
+    return new SyncResponse(syncCount, failCount, bytesTransferred, status.getStatus());
+  }
+
+  private void replicateSubset(Filter filter) {
+    for (Result metacardResult : getMetacards(filter)) {
       if (canceled) {
-        break;
+        return;
       }
       mcard = metacardResult.getMetacard();
       existingReplicationItem = persistentStore.getItem(mcard.getId(), sourceName, destinationName);
@@ -153,7 +170,7 @@ class SyncHelper {
         if (causedByConnectionLoss(e)) {
           logConnectionLoss();
           status.setStatus(Status.CONNECTION_LOST);
-          return new SyncResponse(syncCount, failCount, bytesTransferred, status.getStatus());
+          return;
         } else {
           recordItemFailure(e);
         }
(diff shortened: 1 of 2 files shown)


replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
```diff
@@ -139,7 +139,24 @@ public SyncHelper(
...
-    for (Result metacardResult : getMetacardChangeSet()) {
+    // Failed items and new items are replicated separately due to a bug in DDF that was only fixed
+    // recently
+    List<Filter> failedItemFilters = createFailedMetacardFilters();
+    if (!failedItemFilters.isEmpty()) {
+      replicateSubset(builder.anyOf(failedItemFilters));
+    }
+
+    // if the connection was lost while replicating failed items don't continue replicating
+    if (!status.getStatus().equals(Status.CONNECTION_LOST)) {
```
paouelle: ❓ Should we skip this if replication was cancelled? Should it actually be `!= Status.SUCCESS`? The logic would be easier to read if the status would start at SUCCESS and then downgrade from that where it fails or is checked to be cancelled. Then finally aborted here if it is no longer success.

✏️ BTW, if Status is an enum then you should simply use `==` or `!=` and not `equals()`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
```diff
@@ -139,7 +139,24 @@ public SyncHelper(
...
+      replicateSubset(buildChangeSetFilter());
+    }
+
+    status.setStatus(canceled ? Status.CANCELED : Status.SUCCESS);
+    return new SyncResponse(syncCount, failCount, bytesTransferred, status.getStatus());
+  }
+
+  private void replicateSubset(Filter filter) {
+    for (Result metacardResult : getMetacards(filter)) {
       if (canceled) {
```
paouelle: ❓ How about setting the status to `CANCELED`?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
```diff
@@ -153,7 +170,7 @@ public SyncResponse sync() {
         if (causedByConnectionLoss(e)) {
           logConnectionLoss();
           status.setStatus(Status.CONNECTION_LOST);
-          return new SyncResponse(syncCount, failCount, bytesTransferred, status.getStatus());
+          return;
```
paouelle: ✏️ Either break here or return above to keep it consistant

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
kcover: If we skip it if it is cancelled that would save us a query, I like that. 👍 The expected status would actually be PULL_IN_PROGRESS or PUSH_IN_PROGRESS and I think we use that to fetch the status for the UI so I didn't want to mess with that. Also will change to use `!=`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
kcover: We set the status to canceled almost immediately after this break. I could set it here if you find that more readable.

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/SyncHelperTest.java
```diff
@@ -164,4 +216,11 @@ public void isCanceled() {
     helper.cancel();
     assertThat(helper.isCanceled(), is(true));
   }
+
+  private void setConfigDefaults() {
```
peterhuffer: ✏️ nit: `mockConfig`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/SyncHelper.java
paouelle: By doing it here, it is closer to where the decision is made as opposed to be somewhere else along the flow.

PR #124, opened 2019-06-26
REP-119 Fix failure retry query bug

### LINKED ISSUE #119: When failure retries are included in csw query no new items are returned

kcover:
#### What does this PR do?
This commit fixes a bug where no new items are returned when failure
retries are included in a csw query to DDF

Test steps are too long to list here and will likely change soon 😄. See me for information.
  
Fixes #119 

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
+++ b/adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
@@ -53,6 +53,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status.Family;
@@ -261,13 +262,14 @@ public class DdfNodeAdapter implements NodeAdapter {
 
   @Override
   public QueryResponse query(QueryRequest queryRequest) {
+    // Failed items and new items are queried separately due to a bug in DDF that was only fixed
+    // recently
     Iterable<Metadata> results =
         ResultIterable.resultIterable(
             this::getRecords,
-            new QueryRequestImpl(
-                createQueryFilter(queryRequest),
-                queryRequest.getStartIndex(),
-                queryRequest.getPageSize()));
+            createDdfFailedItemQueryRequest(queryRequest),
+            createDdfQueryRequest(queryRequest));
+
     return new QueryResponseImpl(results);
   }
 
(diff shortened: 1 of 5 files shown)


adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
```diff
@@ -261,13 +261,19 @@ public String getSystemName() {
...
-                queryRequest.getStartIndex(),
-                queryRequest.getPageSize()));
+    Iterable<Metadata> results;
+    // Failed items and new items are replicated separately due to a bug in DDF that was only fixed
+    // recently
+    if (queryRequest.getFailedItemIds().isEmpty()) {
+      results =
+          ResultIterable.resultIterable(this::getRecords, createDdfQueryRequest(queryRequest));
+    } else {
+      results =
```
clockard: if we made the createDdfFailedItemsQueryRequest just return null if there were no failed items and then had resultIterable just filter out null values it would simplify this. Not sure that it's worth it but just a thought.

kcover: I made the changes you suggested.

adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/ResultIterable.java
```diff
@@ -179,9 +197,19 @@ private boolean isDistinctResult(@Nullable Metadata result) {
       return result != null && (result.getId() == null || !foundIds.contains(result.getId()));
     }
 
-    private void copyQueryRequestAndQuery(QueryRequest request, int index) {
+    private void nextRequestOrSignalFinished() {
```
paouelle: :question: Are we concerned by the recursion introduced between `fetchNextResults()` and `nextRequestOrSignalFinished()`? Should the logic be changed to avoid that?

kcover: I asked @clockard about that and we seemed to agree that It shouldn't be something we need to worry about. It should only be us using this class and I would hope none of us would put a few thousand query requests into the result iterable.

clockard:
Heroed ✅

PR #130, opened 2019-06-27
REP-129 Fix initial replication to Ion bug

### LINKED ISSUE #129: First attempt at replicating an item with Ion fails
**Describe the bug**
The first attempt to replicate an item to Ion always fails.

**Affects version**
0.3.0

**To Reproduce**
- Setup replication with a ddf based system as the source and ion as the destination
- Observe as the ingest fails and an error with a 500 status code is returned in the logs

**Expected behavior**
The item should replicate on the first attempt.

kcover:
### What does this PR do?
*Fixes a bug where initial attempts to replicate a resource to Ion, that
wasn't a text resource, would fail because the file size sent to Ion was
-1.

*Modifies the IonNodeAdapter to consider any create resource response
with a status code in the 200 range to be successful.

- Set up a replication to Ion which will replicate an image.
- Verify that the replication is successful on the first attempt.

Fixes #129 

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/adapters/ion-adapter/src/main/java/com/connexta/replication/adapters/ion/IonNodeAdapter.java
+++ b/adapters/ion-adapter/src/main/java/com/connexta/replication/adapters/ion/IonNodeAdapter.java
@@ -131,7 +131,7 @@ public class IonNodeAdapter implements NodeAdapter {
         ResponseEntity<String> response =
             restOps.exchange(
                 this.ionUrl.toString() + "/ingest", HttpMethod.POST, requestEntity, String.class);
-        if (response.getStatusCodeValue() != 202) {
+        if (!response.getStatusCode().is2xxSuccessful()) {
           LOGGER.debug(
               "Failed to replicate {}. Message: {}", resource.getName(), response.getBody());
           success = false;
--- a/adapters/ion-adapter/src/test/java/com/connexta/replication/adapters/ion/IonNodeAdapterTest.java
+++ b/adapters/ion-adapter/src/test/java/com/connexta/replication/adapters/ion/IonNodeAdapterTest.java
@@ -80,22 +80,33 @@ public class IonNodeAdapterTest {
   }
 
   @Test
-  public void createResourceFailed401() {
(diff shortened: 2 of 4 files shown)


kcover:
just added a test to maintain the code coverage we had.

adapters/ion-adapter/src/main/java/com/connexta/replication/adapters/ion/IonNodeAdapter.java
```diff
@@ -131,7 +131,7 @@ public boolean createResource(CreateStorageRequest createStorageRequest) {
         ResponseEntity<String> response =
             restOps.exchange(
                 this.ionUrl.toString() + "/ingest", HttpMethod.POST, requestEntity, String.class);
-        if (response.getStatusCodeValue() != 202) {
+        if (!response.getStatusCode().is2xxSuccessful()) {
```
paouelle: 👍 Much better

adapters/ion-adapter/src/test/java/com/connexta/replication/adapters/ion/IonNodeAdapterTest.java
```diff
@@ -80,22 +80,33 @@ public void createResource() {
   }
 
   @Test
-  public void createResourceFailed401() {
+  public void createResource200() {
```
paouelle: ✏️ Should probably be adding another test with another 2xx code to properly verify your change above.

kcover: There is a test above this where we respond with a 202. Or did you mean 2 2xx tests isn't enough?

paouelle: I missed it.

clockard:
Heroed ✅

PR #134, opened 2019-07-01
REP-133 Make connection timeouts configurable

### LINKED ISSUE #133: Connections to DDF timeout for DFF instances that take longer than 60 seconds to respond
**Describe the bug**
It has come up that the default 60 second response time for querying DDF systems is timing out for very slow response systems. A mechanism needs to be added so that this timeout can be increased.

**Affects version**
0.2.2, 0.3.0

kcover:
#### What does this PR do?
Makes the connect and receive timeouts for contacting DDF configurable
via system property.

-Install and configure a DDF for replication.
-Follow the added documentation to configure the replication timeouts
-Install replication
-Set the log level for `org.codice.ditto.replication.api.impl.ReplicatorStoreFactoryImpl` to DEBUG
-Create and run a replication
-Watch for the timeouts to be logged and verify they are the expected values

Fixes #133 

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/docs/src/main/resources/content/replication-configuring.adoc
+++ b/docs/src/main/resources/content/replication-configuring.adoc
@@ -105,3 +105,17 @@ of the replication configuration regardless of the results of the above options.
 
 To view a list of all currently saved replication configurations, simply execute `replication:config-list`
 in the ${command-console}.
+
+=== Setting Timeouts
+
+By default, requests made by the replication feature have a connection timeout of 30 seconds and a receive
+timeout of 60 seconds. If your network is slow, or the ${branding} you are trying to replicate with is
+slow to respond, these timeouts can be increased to allow more time for transferring information.
+Here's how you can set these timeouts:
+
+. Navigate to the `<${branding}-home>/etc` directory.
+. Open the `custom.system.properties` file.
+. Create an entry specifying the connection timeout in seconds ex: `replication.connection.timeout=30`.
+. Create an entry specifying the receive timeout in seconds ex: `replication.receive.timeout=60`.
+. Save the file.
+. Restart the ${branding} if it is running.
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorStoreFactoryImpl.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorStoreFactoryImpl.java
@@ -15,6 +15,7 @@ package org.codice.ditto.replication.api.impl;
 
 import static org.codice.ddf.spatial.ogc.csw.catalog.common.CswAxisOrder.LON_LAT;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.thoughtworks.xstream.converters.Converter;
 import ddf.catalog.CatalogFramework;
 import ddf.catalog.filter.FilterAdapter;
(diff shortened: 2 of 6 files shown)


docs/src/main/resources/content/replication-configuring.adoc
```diff
@@ -104,4 +104,15 @@ of the replication configuration regardless of the results of the above options.
...
\ No newline at end of file
+in the ${command-console}.
+
+=== Setting Timeouts
+
+Here's how you can set the timeouts for requests to a ${branding}:
+
+. Navigate to the `<${branding}-home>/etc` directory.
+. Open the `custom.system.properties` file.
+. Create an entry specifying the connection timeout in seconds ex: `replication.connection.timeout=30`.
```
paouelle: ❓ Should we document the defaults?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorStoreFactoryImpl.java
```diff
@@ -45,9 +53,9 @@
 
   private static final Integer POLL_INTERVAL_MINUTES = 5;
 
-  private static final Integer CONNECTION_TIMEOUT = 30000;
+  private static final Integer DEFAULT_CONNECTION_TIMEOUT = 30;
```
paouelle: ✏️ Please documents that they are in seconds or add something like `_IN_SECS` to the name.

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/ReplicatorStoreFactoryImplTest.java
```diff
@@ -81,10 +89,54 @@ public void setUp() throws Exception {
     factory.setCatalogFramework(framework);
   }
 
+  @After
+  public void tearDown() throws Exception {
+    System.clearProperty("replication.connection.timeout");
```
paouelle: ✏️ Please use the `RestoreSystemProperties` JUnit rule or its annotation instead

docs/src/main/resources/content/replication-configuring.adoc
```diff
@@ -104,4 +104,15 @@ of the replication configuration regardless of the results of the above options.
 === Viewing Replication Configurations
 
 To view a list of all currently saved replication configurations, simply execute `replication:config-list`
-in the ${command-console}.
\ No newline at end of file
+in the ${command-console}.
+
+=== Setting Timeouts
+
+Here's how you can set the timeouts for requests to a ${branding}:
```
peterhuffer: ❓ Should we include a why? For example, `If your network is slow, increasing the client timeouts for replication can be increased to allow more time for transferring resources.`

docs/src/main/resources/content/replication-configuring.adoc
```diff
@@ -104,4 +104,15 @@ of the replication configuration regardless of the results of the above options.
...
+
+. Navigate to the `<${branding}-home>/etc` directory.
+. Open the `custom.system.properties` file.
+. Create an entry specifying the connection timeout in seconds ex: `replication.connection.timeout=30`.
+. Create an entry specifying the receive timeout in seconds ex: `replication.receive.timeout=60`.
+. Save the file.
+. Restart the ${branding}.
```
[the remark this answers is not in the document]
kcover: I wasn't sure if karaf needed a restart to reread the system properties in custom.system.properties when I wrote this. I will double check and modify the docs if needed.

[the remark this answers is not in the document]
kcover: I had to restart to get the changes to take effect, we shouldn't need to restart when using spring properties however 😄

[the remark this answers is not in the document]
paouelle: I believe Karaf only reads properties at startup; so changing the file won't have any effect unless you restart the system.

peterhuffer:
Successful hero ✔️

PR #141, opened 2019-07-12
REP-136 Allow the user to use a pre-saved query for replication

### LINKED ISSUE #136: As a user, I'd like to create replications with saved queries
**Description**
Trying to create a query with raw cql is difficult. Add a component to the UI that will allow the user to select a query that is already saved on the system when creating a replication.

Fix Version: 0.2.3

kcover:
#### What does this PR do?
Allows the user to select a pre-saved query when creating a replication so they can use its filter.

install replication, create some queries and test out the new ui component. Get creative and try to break it 😉 . You should also save with one of the created queries and make sure replication will actually work with that filter.

Fixes #136 

Capture
Selecting
selected

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationFieldProvider.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationFieldProvider.java
@@ -17,6 +17,7 @@ import com.google.common.collect.ImmutableList;
 import java.util.List;
 import org.codice.ddf.admin.api.fields.FunctionField;
 import org.codice.ddf.admin.common.fields.base.function.BaseFieldProvider;
+import org.codice.ditto.replication.admin.query.queries.GetQueries;
 import org.codice.ditto.replication.admin.query.replications.discover.GetReplications;
 import org.codice.ditto.replication.admin.query.replications.persist.CancelReplication;
 import org.codice.ditto.replication.admin.query.replications.persist.CreateReplication;
@@ -30,39 +31,47 @@ import org.codice.ditto.replication.admin.query.sites.persist.UpdateReplicationS
 import org.codice.ditto.replication.admin.query.status.persist.UpdateReplicationStats;
 import org.codice.ditto.replication.admin.query.ui.GetUiConfig;
 
+/** Provides the graphql endpoint with the functionality we would like to expose. */
 public class ReplicationFieldProvider extends BaseFieldProvider {
 
-  public static final String DEFAULT_FIELD_NAME = "replication";
+  private static final String DEFAULT_FIELD_NAME = "replication";
 
-  public static final String TYPE_NAME = "Replication";
+  private static final String TYPE_NAME = "Replication";
 
-  public static final String DESCRIPTION =
+  private static final String DESCRIPTION =
       "Provides methods for querying and modifying replication configurations and sites.";
 
-  private GetReplications getReplications;
+  private final GetReplications getReplications;
 
-  private GetReplicationSites getReplicationSites;
+  private final GetReplicationSites getReplicationSites;
 
-  private CreateReplication createReplication;
+  private final CreateReplication createReplication;
 
-  private UpdateReplication updateReplication;
+  private final UpdateReplication updateReplication;
 
-  private DeleteReplication deleteReplication;
+  private final DeleteReplication deleteReplication;
 
-  private CreateReplicationSite createReplicationSite;
+  private final CreateReplicationSite createReplicationSite;
 
-  private UpdateReplicationSite updateReplicationSite;
+  private final UpdateReplicationSite updateReplicationSite;
 
-  private DeleteReplicationSite deleteReplicationSite;
+  private final DeleteReplicationSite deleteReplicationSite;
 
-  private CancelReplication cancelReplication;
+  private final CancelReplication cancelReplication;
 
-  private SuspendReplication suspendReplication;
+  private final SuspendReplication suspendReplication;
 
-  private GetUiConfig getUiConfig;
+  private final GetUiConfig getUiConfig;
 
-  private UpdateReplicationStats updateReplicationStats;
+  private final UpdateReplicationStats updateReplicationStats;
 
+  private final GetQueries getQueries;
+
+  /**
+   * Creates a new ReplicationFieldProvider. The parameters all extend {@link
+   * org.codice.ddf.admin.common.fields.base.BaseFunctionField} and will provide functionality which
+   * we will expose through the graphql endpoint.
+   */
   public ReplicationFieldProvider(
       GetReplications getReplications,
       GetReplicationSites getReplicationSites,
@@ -75,7 +84,8 @@ public class ReplicationFieldProvider extends BaseFieldProvider {
       UpdateReplicationSite updateReplicationSite,
       DeleteReplicationSite deleteReplicationSite,
       GetUiConfig getUiConfig,
-      UpdateReplicationStats updateReplicationStats) {
+      UpdateReplicationStats updateReplicationStats,
+      GetQueries getQueries) {
     super(DEFAULT_FIELD_NAME, TYPE_NAME, DESCRIPTION);
     this.getReplications = getReplications;
     this.getReplicationSites = getReplicationSites;
@@ -89,13 +99,28 @@ public class ReplicationFieldProvider extends BaseFieldProvider {
     this.deleteReplicationSite = deleteReplicationSite;
     this.getUiConfig = getUiConfig;
     this.updateReplicationStats = updateReplicationStats;
+    this.getQueries = getQueries;
   }
 
+  /**
+   * Returns a list of {@link FunctionField}s which, when accessed by the graphql endpoint, will
+   * allow their functionality to be triggered through the graphql endpoint.
+   *
+   * @return a list of {@link FunctionField}s that will allow retrieval of data through the graphql
+   *     endpoint.
+   */
   @Override
   public List<FunctionField> getDiscoveryFunctions() {
-    return ImmutableList.of(getReplications, getReplicationSites, getUiConfig);
+    return ImmutableList.of(getReplications, getReplicationSites, getUiConfig, getQueries);
   }
 
+  /**
+   * Returns a list of {@link FunctionField}s which, when accessed by the graphql endpoint, will
+   * allow their functionality to be triggered through the graphql endpoint.
+   *
+   * @return a list of {@link FunctionField}s that will allow mutation of data through the graphql
+   *     endpoint.
+   */
   @Override
   public List<FunctionField> getMutationFunctions() {
     return ImmutableList.of(
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/GetQueries.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/GetQueries.java
@@ -1,0 +1,102 @@
+package org.codice.ditto.replication.admin.query.queries;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import java.util.List;
+import java.util.Set;
+import org.codice.ddf.admin.api.Field;
+import org.codice.ddf.admin.api.fields.FunctionField;
+import org.codice.ddf.admin.api.fields.ListField;
+import org.codice.ddf.admin.common.fields.base.BaseFunctionField;
+import org.codice.ditto.replication.api.QueryService;
+import org.codice.ditto.replication.api.data.Query;
+
+/**
+ * This is a graphql function field. A call to the graphql endpoint with the given field name will
+ * trigger the performFunction method and present the returned data in the response.
+ */
+public class GetQueries extends BaseFunctionField<ListField<QueryField>> {
+
+  private static final String FIELD_NAME = "queries";
+
+  private static final String DESCRIPTION =
+      "Retrieves all saved queries for the currently logged in user.";
+
+  private static final ListField<QueryField> RETURN_TYPE = new QueryField.ListImpl();
+
+  private final QueryService queryService;
+
+  /**
+   * Creates a new GetQueries function field.
+   *
+   * @param queryService A service which will fetch queries from the catalog framework for us
+   */
+  public GetQueries(QueryService queryService) {
+    super(FIELD_NAME, DESCRIPTION);
+    this.queryService = queryService;
+  }
+
+  /**
+   * Contains all the functionality of the "queries" graphql query. The response of the query will
+   * contain the information in the {@link QueryField.ListImpl}.
+   *
+   * @return the result of a query for saved user queries in the form of a {@link
+   *     QueryField.ListImpl}
+   * @throws org.codice.ditto.replication.api.QueryException if an exception is thrown when querying
+   *     the catalog.
+   */
+  @Override
+  public ListField<QueryField> performFunction() {
+    ListField<QueryField> queryFields = new QueryField.ListImpl();
+    queryService.queries().map(this::getFieldForQuery).forEach(queryFields::add);
+    return queryFields;
+  }
+
+  private QueryField getFieldForQuery(Query query) {
+    QueryField queryField = new QueryField();
+    queryField.title(query.getTitle());
+    queryField.cql(query.getCql());
+    return queryField;
+  }
+
+  /**
+   * Returns the error codes this function field can return
+   *
+   * @return the error codes this function field can return
+   */
+  @Override
+  public Set<String> getFunctionErrorCodes() {
+    return ImmutableSet.of();
+  }
+
+  /**
+   * Returns the arguments to the {@link BaseFunctionField#performFunction()} method of this
+   * function field
+   *
+   * @return arguments to {@link BaseFunctionField#performFunction()} for this function field
+   */
+  @Override
+  public List<Field> getArguments() {
+    return ImmutableList.of();
+  }
+
+  /**
+   * Returns the return type of this function field
+   *
+   * @return the return type of this function field
+   */
+  @Override
+  public ListField<QueryField> getReturnType() {
+    return RETURN_TYPE;
+  }
+
+  /**
+   * Returns a new instance of this function field
+   *
+   * @return a new instance of this function field
+   */
+  @Override
+  public FunctionField<ListField<QueryField>> newInstance() {
+    return new GetQueries(queryService);
+  }
+}
--- a/admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/QueryField.java
+++ b/admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/QueryField.java
@@ -1,0 +1,125 @@
+package org.codice.ditto.replication.admin.query.queries;
+
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+import java.util.concurrent.Callable;
+import org.codice.ddf.admin.api.Field;
+import org.codice.ddf.admin.common.fields.base.BaseListField;
+import org.codice.ddf.admin.common.fields.base.BaseObjectField;
+import org.codice.ddf.admin.common.fields.base.scalar.StringField;
+
+/** A data structure representing a saved query, created for use with the graphql endpoint. */
+public class QueryField extends BaseObjectField {
+  private static final String DEFAULT_FIELD_NAME = "query";
+
+  private static final String FIELD_TYPE_NAME = "Query";
+
+  private static final String DESCRIPTION =
+      "A query which contains a filter that can be extracted and used in a replication.";
+
+  private StringField title;
+
+  private StringField cql;
+
+  /**
+   * Creates a new QueryField used to make query information available through the graphql endpoint.
+   */
+  public QueryField() {
+    super(DEFAULT_FIELD_NAME, FIELD_TYPE_NAME, DESCRIPTION);
+    this.title = new StringField("title");
+    this.cql = new StringField("cql");
+  }
+
+  /**
+   * Sets the title of this QueryField to the given title and returns the entire field.
+   *
+   * @param title The title to give to this QueryField
+   * @return this QueryField
+   */
+  public QueryField title(String title) {
+    this.title.setValue(title);
+    return this;
+  }
+
+  /**
+   * Sets the cql of this QueryField to the given cql and returns the entire field.
+   *
+   * @param cql The cql to set for this QueryField
+   * @return this QueryField
+   */
+  public QueryField cql(String cql) {
+    this.cql.setValue(cql);
+    return this;
+  }
+
+  /**
+   * Returns the title of the query as a {@link String}
+   *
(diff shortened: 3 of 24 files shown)


replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+      Metacard mcard = result.getMetacard();
+      queries.addAll(getQueries(mcard));
+    }
+    return queries;
+  }
+
+  public List<Query> getQueries(Metacard metacard) {
+    List<Query> queries = new ArrayList<>();
+    List<Metacard> metacards =
+        getQueryStrings(metacard).stream().map(this::xmlToMetacard).collect(Collectors.toList());
```
paouelle: ❓ Why collect just to iterate that temporary collection? Just continue streaming to perform what you have in the loop below. Suggestion would be to create a private method that accepts a `Metacard` and returns a `org.codice.ditto.replication.api.impl.data.QueryImpl` which would make the stream much more readable and this function much simpler.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+      String cql = (String) mcard.getAttribute("cql").getValue();
+      queries.add(new org.codice.ditto.replication.api.impl.data.QueryImpl(title, cql));
+    }
+    return queries;
+  }
+
+  private List<String> getQueryStrings(Metacard metacard) {
+    Attribute attribute = metacard.getAttribute("queries");
+    if (attribute != null) {
+      return attribute.getValues().stream().map(String::valueOf).collect(Collectors.toList());
```
paouelle: Nice

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+    return Collections.emptyList();
+  }
+
+  public Metacard xmlToMetacard(String xml) {
```
paouelle: ✏️ Should be `private`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/QueryImpl.java
```diff
@@ -0,0 +1,35 @@
+package org.codice.ditto.replication.api.impl.data;
+
+import org.codice.ditto.replication.api.data.Query;
+
+public class QueryImpl implements Query {
```
paouelle: ❗️ javadoc

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/QueryImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private String title;
```
paouelle: ❗️ Should be annotated with `@Nullable`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/QueryImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private String cql;
```
paouelle: ❗️ Should be annotated with `@Nullable`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/QueryImpl.java
```diff
@@ -0,0 +1,35 @@
...
+  public QueryImpl() {
+    this(null, null);
+  }
+
+  public QueryImpl(String title, String cql) {
+    this.title = title;
+    this.cql = cql;
+  }
+
+  public String getTitle() {
```
paouelle: ❗️ Should be annotated with `@Nullable`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/QueryImpl.java
```diff
@@ -0,0 +1,35 @@
...
+    return title;
+  }
+
+  public void setTitle(String title) {
+    this.title = title;
+  }
+
+  public String getCql() {
```
paouelle: ❗️ Should be annotated with `@Nullable`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/QueryImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+  private String title;
+
+  private String cql;
+
+  public QueryImpl() {
+    this(null, null);
+  }
+
+  public QueryImpl(String title, String cql) {
```
paouelle: ✏️ javadoc
❗️ Both parameters should be annotated with `@Nullable`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/QueryImpl.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.codice.ditto.replication.api.data.Query;
+
+public class QueryImpl implements Query {
+
+  private String title;
+
+  private String cql;
+
+  public QueryImpl() {
```
paouelle: ✏️ javadoc

replication-api/src/main/java/org/codice/ditto/replication/api/QueryService.java
```diff
@@ -0,0 +1,9 @@
+package org.codice.ditto.replication.api;
+
+import java.util.List;
+import org.codice.ditto.replication.api.data.Query;
+
+public interface QueryService {
```
paouelle: ❗️ javadoc

replication-api/src/main/java/org/codice/ditto/replication/api/QueryService.java
```diff
@@ -0,0 +1,9 @@
...
+
+  public List<Query> getQueries();
```
paouelle: ❗️ Javadoc
❓ Could we return a `Stream` instead of a collection?

replication-api/src/main/java/org/codice/ditto/replication/api/data/Query.java
```diff
@@ -0,0 +1,8 @@
+package org.codice.ditto.replication.api.data;
+
+public interface Query {
```
paouelle: ❗️ javadoc

replication-api/src/main/java/org/codice/ditto/replication/api/data/Query.java
```diff
@@ -0,0 +1,8 @@
...
+
+  public String getTitle();
```
paouelle: ❗️ javadoc. Should be annotated with `@Nullable` based on the implementation above.

replication-api/src/main/java/org/codice/ditto/replication/api/data/Query.java
```diff
@@ -0,0 +1,8 @@
...
+
+  public String getCql();
```
paouelle: ❗️ javadoc. Should be annotated with `@Nullable` based on the implementation above.

kcover:
Confirmed that only the admin's queries are retrieved for selection.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationFieldProvider.java
```diff
@@ -63,6 +64,8 @@
 
   private UpdateReplicationStats updateReplicationStats;
 
+  private GetQueries getQueries;
```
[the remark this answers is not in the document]
kcover: sure.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/GetQueries.java
```diff
@@ -0,0 +1,63 @@
...
+      queryField.cql(query.getCql());
+      queryFields.add(queryField);
+    }
+
+    return queryFields;
+  }
+
+  @Override
+  public Set<String> getFunctionErrorCodes() {
+    return ImmutableSet.of();
```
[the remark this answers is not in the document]
kcover: Just to keep it consistent with the other Field Classes. Also makes it a hair easier to add in elements if we decide we need to.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/QueryField.java
```diff
@@ -0,0 +1,73 @@
...
+
+  private StringField cql;
+
+  public QueryField() {
+    super(DEFAULT_FIELD_NAME, FIELD_TYPE_NAME, DESCRIPTION);
+    this.title = new StringField("title");
+    this.cql = new StringField("cql");
+  }
+
+  public QueryField title(String title) {
```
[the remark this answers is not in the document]
kcover: Is there a reason the getters should have javadoc?

admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/GetQueries.java
peterhuffer: ✏️ We can use Java's `Set.of()` now, which returns an immutable set.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/GetQueries.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Override
+  public List<Field> getArguments() {
+    return ImmutableList.of();
```
peterhuffer: ✏️ We can use Java's `List.of()` now, which returns an immutable list.

replication-api/src/main/java/org/codice/ditto/replication/api/QueryService.java
kcover: yes.

replication-api/src/main/java/org/codice/ditto/replication/api/QueryService.java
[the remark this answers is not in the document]
paouelle: All classes and public methods / attributes should be documented properly. Sometimes that can be very simple in deed.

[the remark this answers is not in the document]
kcover: Everything I've read agrees with you but nothing explains the value in adding javadoc to a simple method like a getter. It seems like it's just more text for the developer to skip over. So my question is, why should they have javadoc? I could be missing something because I very rarely look at javadoc.

ui/src/main/webapp/components/replications/QuerySelector.js
```diff
@@ -0,0 +1,61 @@
...
+    const queryList = Immutable.List(queries)
+    return queryList.map(query => ({ value: query.cql, label: query.title }))
+  }
+
+  render() {
+    const { onChange } = this.props
+    return (
+      <Query query={allQueries}>
+        {({ loading, error, data, refetch }) => {
+          if (loading) return <Typography>Loading...</Typography>
```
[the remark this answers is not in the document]
kcover: I Started using the isLoading prop in the select component. I gives us a nice substitute for the menu when It's loading.
Loading

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+   * Provides a stream for retrieving all saved user queries as {@link Query}s.
+   *
+   * @return all the user queries saved on the system
+   */
+  @Override
+  public Stream<Query> queries() {
+    Filter filter = builder.attribute(Core.METACARD_TAGS).is().equalTo().text("workspace");
+    QueryResponse queryResponse;
+    try {
+      queryResponse = framework.query(new QueryRequestImpl(new QueryImpl(filter)));
```
clockard: This query will return a max of 20 workspaces since a page size isn't specified. Should we increase that?

clockard: Do we want to consider using ResultsIterable?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    return queryResponse
+        .getResults()
+        .stream()
+        .map(Result::getMetacard)
+        .flatMap(this::getQueryXmlStrings)
+        .map(this::xmlToMetacard)
+        .map(this::createQueryFromMetacard);
+  }
+
+  private Stream<String> getQueryXmlStrings(Metacard metacard) {
```
paouelle: ✏️ By conventions methods that returns `Stream` should not be prefixed with `get`

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
kcover: I've increased the page size to 100, which I think should be more than a reasonable number of workspaces for a user to have.

kcover:
test container failed to come up for some reason. build now

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * Provides a stream for retrieving all saved user queries as {@link Query}s.
+   *
+   * @return all the user queries saved on the system
+   */
+  @Override
+  public Stream<Query> queries() {
+    Filter filter = builder.attribute(Core.METACARD_TAGS).is().equalTo().text("workspace");
+    QueryImpl query = new QueryImpl(filter);
+    query.setPageSize(100);
```
peterhuffer: ❓ Will we only ever return the first 100 workspaces? Could that be a problem?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
```diff
@@ -0,0 +1,99 @@
...
+    QueryResponse queryResponse;
+    try {
+      queryResponse = framework.query(new QueryRequestImpl(query));
+    } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
+      throw new ReplicationException("Failed to retrieve Query metacards.", e);
```
peterhuffer: ❓ Should we really be using this exception for this?

replication-api/src/main/java/org/codice/ditto/replication/api/QueryService.java
```diff
@@ -0,0 +1,15 @@
...
+import java.util.stream.Stream;
+import org.codice.ditto.replication.api.data.Query;
+
+/** A service which fetches saved user {@link Query}s. */
+public interface QueryService {
+
+  /**
+   * Provides a stream for retrieving all saved user queries as {@link Query}s.
+   *
+   * @return all the user queries saved on the system
```
peterhuffer: ✏️ This should mention throwing ReplicationException.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/GetQueries.java
```diff
@@ -0,0 +1,100 @@
...
+   * Contains all the functionality of the "queries" graphql query. The response of the query will
+   * contain the information in the {@link QueryField.ListImpl}.
+   *
+   * @return the result of a query for saved user queries in the form of a {@link
+   *     QueryField.ListImpl}
+   */
+  @Override
+  public ListField<QueryField> performFunction() {
+    ListField<QueryField> queryFields = new QueryField.ListImpl();
+    queryService.queries().map(this::getFieldForQuery).forEach(queryFields::add);
```
peterhuffer: ❓ Should we catch the ReplicationException?

peterhuffer:
❗️ There should be an itests for the Queries function.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
kcover: Yea we only return the first 100 workspaces. I think that it's reasonable to expect one user to not have more that 100 workspaces but could see this being an issue if, say, multiple people shared one set of credentials. Then there's a good chance they will have more that 100. We could return a resultIterable but then we would need to incorporate paging into the query selector. Or we could only fetch queries based on what the user types into the search box, but I think it would be difficult to search based on query titles since they are saved as metacard xml on workspace metacards.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
paouelle: If not then we should declare it can be thrown

paouelle: And that would need to be done in the interface as we shouldn't be throwing out exceptions in an implementation that are not documented in the interface.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
paouelle: I thought there was a notion of shared workspaces. If that is the case and if the list be greater than 100, how do we know that the first 100 are the right one to display?

clockard: I feel like 98% of the time this will be sufficient. I don't think we should hold this pr up for this reason. If needed we can iterate on some kind of paging. We could also add a sort to our query so the most recent 100 workspaces are returned with would probably make it work 99% of the time. (like how I am just making up %)

paouelle: I would go for the last thing you suggested if it is easy.

replication-api/src/main/java/org/codice/ditto/replication/api/QueryService.java
paouelle: I do agree with your previous comment @peterhuffer that using `ReplicationException` here is weird since this is a query interface and not replication specific. 

Even if it had been replication specific, I am never a fan of re-using a base and generic exception like `ReplicationException` to wrap everything and anything. Sometimes, it is much better to define subclasses of that one to be a bit more specific.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
kcover: I'll add a sort based on the modified date. That should help.

admin/query/src/main/java/org/codice/ditto/replication/admin/query/ReplicationFieldProvider.java
```diff
@@ -30,39 +31,47 @@
...
-  private UpdateReplicationStats updateReplicationStats;
+  private final UpdateReplicationStats updateReplicationStats;
 
+  private final GetQueries getQueries;
+
+  /**
+   * Creates a new ReplicationFieldProvider. The parameters all extend {@link
+   * org.codice.ddf.admin.common.fields.base.BaseFunctionField} and will provide functionality which
+   * we will expose through the graphql endpoint.
+   */
```
paouelle: ✏️ javadoc the parameters

admin/query/src/main/java/org/codice/ditto/replication/admin/query/queries/GetQueries.java
paouelle: If not then you need to document it can be thrown out

kcover: I think it would only make sense to catch it if we wanted to return an error message, and we only return an error message if there's something that can be done outside of the system to correct the issue. Since any issue that would cause the exception would have to be internal, I think we should just let the exception bubble out and have the graphql enpdoint return with the "internal server error" message.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
kcover: I created a QueryException and it throws that now. I also added documentation to all the methods it could bubble out of.

distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
```diff
@@ -52,23 +64,52 @@
 )
 @ServiceAdmin
 @ConfigurationAdmin
+// keeps solr from changing the password so we can use "admin:admin" with basic auth to contact solr
+@SetSystemProperty(key = "solr.attemptAutoPasswordChange", value = "false")
```
paouelle: ✏️ That is probably something that should be added to the `InstallDistributionExtension` class in DDF

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
```diff
@@ -52,17 +57,19 @@ public QueryServiceImpl(
...
     Filter filter = builder.attribute(Core.METACARD_TAGS).is().equalTo().text("workspace");
-    QueryImpl query = new QueryImpl(filter);
-    query.setPageSize(100);
+    PropertyName sortProperty = new PropertyNameImpl(Core.METACARD_MODIFIED);
+    SortBy sortPolicy = new SortByImpl(sortProperty, SortOrder.DESCENDING);
+    QueryImpl query = new QueryImpl(filter, 1, 100, sortPolicy, false, 0L);
     QueryResponse queryResponse;
     try {
       queryResponse = framework.query(new QueryRequestImpl(query));
     } catch (UnsupportedQueryException | SourceUnavailableException | FederationException e) {
```
paouelle: ❓ Can `UnsupportedQueryException` actually happen in our case?

kcover: I don't think so. Since I'm querying the local catalog I don't think any of these exceptions will occur.

clockard:
Heroed ✅

PR #144, opened 2019-07-19
Adds support for metrics

peterhuffer:
#### What does this PR do?
- Add support for prometheus metrics. Preferred UI is Grafana
- Removes ReplicationStatus and moves parts of its configuration to ReplicatorConfigs and ReplicationItems
- Adds Action (CREATE|UPDATE|DELETE) to ReplicationItems.
- Make ReplicationItems immutable
- Failures are always retried now

#### Who is reviewing it (please choose AT LEAST two reviewers that need to approve the PR before it can get merged)?
@paouelle @clockard @kcover @oconnormi 

#### How should this be tested? (List steps with links to updated documentation)
1. Download and install prometheus, and configure it to point at replication https://prometheus.io/docs/introduction/first_steps/
2. Download and install Grafana and configure it to point to prometheus
3. Run a replication configuration and observe the stats (see me for a grafana dashboard)
4. Verify creates, updates and deletes

#### What are the relevant tickets?
Fixes #

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests
- [ ] Update / Add Integration Tests

PR #147, opened 2019-07-25
REP-145 Create black box tests

### LINKED ISSUE #145: Create black box integration tests
**Describe the task**
Create integration tests which will stand up replication in a container and let us perform tests against it.

kcover:
#### What does this PR do?
Adds black box tests that utilize external DDF nodes to test replication.

CI Build

Fixes #145 

#### Checklist:
- [x] Documentation Updated
- [ ] Update / Add Unit Tests


distributions/test/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+
+        <!-- TODO: should these be in the top level pom? -->
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
```
kcover: I was wondering why the docker container logger wasn't working and I noticed a logback exception that my eyes had just been skipping over. Adding these dependencies fixed both issues and I was wondering if other modules might be missing these? Should I just put them in the top level pom?

replication-api-impl/src/main/java/com/connexta/ion/replication/api/impl/ReplicatorRunner.java
```diff
@@ -67,8 +67,7 @@ public ReplicatorRunner(
   }
 
   public void init() {
-    scheduledExecutor.scheduleAtFixedRate(
-        this::scheduleReplication, STARTUP_DELAY, period, TimeUnit.SECONDS);
+    scheduledExecutor.scheduleAtFixedRate(this::scheduleReplication, 0, period, TimeUnit.SECONDS);
```
kcover: I don't think we need a delay anymore. I think we're only waiting on solr to come up and spring-data-solr should retry any failed connection attempts.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.concurrent.TimeUnit;
+import org.junit.After;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.output.Slf4jLogConsumer;
+
+// TODO: why does this need to end in "test" for the tests to run?
+public class ReplicationITest extends IntegrationTest {
```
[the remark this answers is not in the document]
peterhuffer: I think we should be able to give the plugin a regex to tell it which files are itests.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/IntegrationTest.java
```diff
@@ -0,0 +1,149 @@
...
+   * @param filePath the path of the file relative to the {@link IntegrationTest} class
+   * @return A String containing the contents of the file
+   */
+  public static String getFileAsText(String filePath) {
+    return getFileAsText(filePath, ImmutableMap.of());
+  }
+
+  /**
+   * Retrieves a file and returns the contents within as a string. The filePath is considered
+   * relative to the class provided. Text in the file can be marked for variable replacement by
```
peterhuffer: ✏️ Javadoc says relative to the class provided, but the IntegrationTest class is used.

distributions/test/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>${junit.version}</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+    <plugins>
```
peterhuffer: ✏️ Formatting looks off.

distributions/test/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.codice.test.thirdparty</groupId>
+            <artifactId>rest-assured</artifactId>
+            <version>${codice-test.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
```
peterhuffer: ✏️ Don't have to do it now, but Java 11 supports a new higher-level HTTP client that we may consider using rather than pulling in a separate library.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/IntegrationTest.java
```diff
@@ -0,0 +1,149 @@
+package com.connexta.ion.replication.distributions.test;
```
peterhuffer: ✏️ Missing licenses.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
```diff
@@ -0,0 +1,144 @@
...
+
+    // create product
+    String id =
+        createProductOnSourceNode(
+            APPLICATION_XML, "data/reuters-xml-sample", ImmutableMap.of("testId", testId));
+    LOGGER.info("Returned ID: {}", id);
+    idsCreated.add(id);
+
+    // create nodes and jobs
+    postToSolrCore(
```
peterhuffer: ❓ Can the configs be created in Before so that all of them can use the configs?

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
```diff
@@ -0,0 +1,144 @@
...
+        given()
+            .relaxedHTTPSValidation()
+            .when()
+            .get(destinationUrl + CATALOG_ENDPOINT + "/" + id)
+            .then()
+            .extract()
+            .body()
+            .xmlPath();
+
+    // TODO: should I verify the replication origins attribute is added?
```
peterhuffer: I don't think this would be necessary. To me this is more of a unit test level assertion.

distributions/test/src/test/resources/application.yml
```diff
@@ -0,0 +1,19 @@
+logging:
+  level:
+    root: INFO
+    org.apache.cxf.interceptor.LoggingOutInterceptor: WARN
+    org.apache.cxf.interceptor.LoggingInInterceptor: WARN
+
+    #turn these up if you suspect issues with solr
```
peterhuffer: ✏️ `suspect` -> `inspect`

distributions/test/pom.xml
peterhuffer: ❓ Is this dependency being used?

distributions/test/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+            <artifactId>httpclient</artifactId>
+            <version>${httpclient.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
```
peterhuffer: ❓ Is this dependency being used?

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
```diff
@@ -0,0 +1,144 @@
...
+    clearAllSolrCores();
+  }
+
+  @Test
+  public void createXmlFile() throws Exception {
+    LOGGER.info("Solr URL: {}", solrUrl);
+    String testId = UUID.randomUUID().toString();
+
+    // create product
+    String id =
```
paouelle: ❓ Where is this cleaned up?

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
```diff
@@ -0,0 +1,144 @@
...
+        .body(getFileAsText(jsonFilePath, varReplacementMap))
+        .when()
+        .post(solrCoreUpdateUrl(core))
+        .then()
+        .statusCode(200);
+  }
+
+  private void deleteFromNodes(String id) {
+    given().relaxedHTTPSValidation().when().delete(sourceUrl + CATALOG_ENDPOINT + "/" + id);
+    given().relaxedHTTPSValidation().when().delete(destinationUrl + CATALOG_ENDPOINT + "/" + id);
```
paouelle: ❓ This cleanup might end up failing on the source and then the destination would not be cleaned up right?

paouelle: ❓ Would it be easier to do a cleanup that simply cleans it all instead of trying to clean specific entries? Kind of like a reset? This would be a reusable test rule.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
paouelle: Or at the minimum, the creation and waiting for it to be there should be done in one utility method that can then be reusable whenever one needs to populate a new entry for testing.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
```diff
@@ -0,0 +1,144 @@
...
+        .relaxedHTTPSValidation()
+        .header(CONTENT_TYPE, contentType)
+        .body(getFileAsText(resourcePath, varReplacementMap))
+        .expect()
+        .log()
+        .headers()
+        .statusCode(201)
+        .when()
+        .post(sourceUrl + CATALOG_ENDPOINT)
+        .getHeader("id");
```
paouelle: ❓ Do we have a guarantee that the product is ready to be queried on the source after this call returns 201?

peterhuffer: I think so since we are querying by ID. DDF will cache uncommitted entries in a near real time cache which will be queried only if it is a query by id.

paouelle: That means we are relying on that specific implementation detail. What is DDF is using SolrCloud? 
Should we be more cautious and safe by making sure it can be queried before we move on?

distributions/test/pom.xml
kcover: codice test rest-assured apparently depends on this. I would think it would contain all of the dependencies it needs already but maybe not?

distributions/test/pom.xml
kcover: I was just using it for Immutable Map which could easily be replaced.

peterhuffer: Ya. Lets use `Map.of()` to avoid this dependency.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/IntegrationTest.java
```diff
@@ -0,0 +1,149 @@
...
+    strSubstitutor.setVariablePrefix(RESOURCE_VARIABLE_DELIMETER);
+    strSubstitutor.setVariableSuffix(RESOURCE_VARIABLE_DELIMETER);
+    String fileContent;
+
+    try {
+      fileContent =
+          IOUtils.toString(
+              IntegrationTest.class.getClassLoader().getResourceAsStream(filePath), "UTF-8");
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to read filepath: " + filePath);
```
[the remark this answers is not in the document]
kcover: AssertionError seems a little misleading here since we're not actually making an assertion. The test will still fail so I would rather keep this as is.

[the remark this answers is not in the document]
paouelle: Sure. Then let's go with either `UncaughtIOException` or `IOError` or `IllegalStateException` since this is something that shouldn't have happen. `RuntimeException` shouldn't be thrown. If static analysis is warning you, and you are forced to put a comment about it, then it means something else should be thrown. Unless you are forced by existing libraries, that declared they threw that (when they should have never done that), just don't.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
```diff
@@ -0,0 +1,144 @@
...
+public class ReplicationITest extends IntegrationTest {
+
+  private static final Logger LOGGER = LoggerFactory.getLogger(ReplicationITest.class);
+
+  private static final Slf4jLogConsumer logConsumer = new Slf4jLogConsumer(LOGGER);
+
+  private Set<String> idsCreated = new HashSet<>();
+
+  @BeforeClass
+  public static void beforeClass() {
```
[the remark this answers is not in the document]
kcover: If a container has an exposed port, Testcontainers will wait for that port to be responive by default. We can also create our own readyness checks. ATM we just use the default readyness check to make sure solr is up. I don't have a readyness check for replication because it should be the last thing we're waiting on. If it doesn't come up in time for one of the tests to pass I don't think it's going to come up at all. If we want to add a readyness check later, we can use the spring-actuator included in #144 to implement a health check for us.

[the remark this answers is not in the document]
paouelle: Being responsive doesn't mean the same as ready.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
kcover: I though a test may need to choose a specific sites or jobs file if it needs a particular set but for now creating them in a Before method should be fine.

paouelle: Personally, I wouldn't create it in `@Before` since the chance of using exactly that in the next test case is slim. I would, however, grouped together the logic into a utility method that you call from here. Or has done in codice-test for config admin, have a way to specify what should be pre-initialized for a given test case before that test case starts. That too me helps separate the pre-conditions (and the corresponding cleanup) from the actual test case

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @After
+  public void cleanUp() {
+    idsCreated.forEach(this::deleteFromNodes);
+    clearAllSolrCores();
+  }
+
+  @Test
+  public void createXmlFile() throws Exception {
```
paouelle: ❓ Should the name be more inline with the intent of the test case? `createXmlFile` seems very short unclear to me

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
kcover: We don't do any assertion here but yea if the source delete throws some kind of exception the destination might not get cleaned up. To delete everything would interfere with testing someone else might be performing with the test instances. To delete just what we created would require a delete by query which isn't possible as far as I know.

paouelle: This test case will have to change as it cannot rely on the pre-existence of another container by hardcoding external URLs as it currently does. I though these were temporary for the `WIP`. Tests must be self contained.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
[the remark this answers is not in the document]
kcover: I looked into it more, apparently we use the fail safe plugin for itests. I Started using that and named the file ReplicationIT. I don't think it gives us any benefit at the moment however, because we're not doing anything in the pre/post-integration-tests phases of the plugin. Testcontainers handles any setup or tear down of the test container instead.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
kcover: I'm having some trouble figuring out how config admin in codice test works, but it seems like we would just be pushing that logic into an annotation instead of a method call? I think I'll stick with making a util method for now.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
kcover: not sure what you mean about hardcoding urls, you can set the source and destination urls using an environment variable but they default to ditto-1 and ditto-2. It turns out I can do a delete by query but then we have to wait for solr to perform a hard commit of what we're deleting. If we delete by id we don't have to wait, we can just remove it right away.

paouelle: My point on hardcoding is that the test expects an external system to be up and running. This is an external dependency that cannot be controlled by the test case and thus will render them brittle and unpredictable. The fact that the URL can be passed in or default to a hardcoded `ditto-1` is the proof of this issue. It might be acceptable for a very short time but it needs to be addressed somehow and very soon.

paouelle: got it for the delete by id. Just put some error handling in the method to not suddenly skip the destination if the source failed for whatever reasons.

distributions/test/src/test/java/com/connexta/ion/replication/distributions/test/ReplicationITest.java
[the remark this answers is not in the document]
kcover: I added a wait strategy to solr that waits for the config core to be ready and I added one to replication that waits for the "started main in ____ seconds" log before proceeding with the tests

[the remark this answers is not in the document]
paouelle: Good. Although we should avoid relying on logs. Could we do or enhance the system such that we would be able to better make that determination as opposed to looking at logs?

distributions/test/src/test/resources/data/reuters-xml-sample
```diff
@@ -0,0 +1,32 @@
+<metacard xmlns="urn:catalog:metacard" xmlns:gml="http://www.opengis.net/gml" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:smil="http://www.w3.org/2001/SMIL20/" xmlns:smillang="http://www.w3.org/2001/SMIL20/Language">
+  <type>ddf.metacard</type>
+  <source>ddf.distribution</source>
+  <string name="title">
+    <value>FED&apos;S JOHNSON SAYS INFLATIONARY FEARS CALMER SINCE LAST DISCOUNT RATE RISE $testId$
```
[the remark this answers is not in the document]
kcover: @clockard suggested just using the generated id as the metacard id but I can't seem to replicate with a filter of "id like '<testid>'" or anything similar. I've kept it the same for now.

distributions/test/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>distributions</artifactId>
+        <groupId>com.connexta.ion.replication.distributions</groupId>
+        <version>0.3.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>itests</artifactId>
```
peterhuffer: ✏️ Name property for this pom.

distributions/test/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>${awaitility.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- TODO: should these be in the top level pom? -->
```
peterhuffer: ✏️ We could probably throw them in the dependency management section of the top level pom, then just put
```
<dependency>
            <groupId>org.slf4j</groupId>
            <artifactId>slf4j-api</artifactId>
            <scope>test</scope>
        </dependency>
```
here.

distributions/test/src/test/resources/jobs.json
```diff
@@ -0,0 +1,13 @@
+ [
+    {
+      "name_txt":"star",
```
peterhuffer: ✏️ Don't forget to update these to match the updated schema.

distributions/test/pom.xml
paouelle: ☝️ That is the standard way to do it. Define version and default scope at the top and reference it with optional overrides when needed.

kcover:
I'll be adding javadoc and a readme for the tests in the next commit.

distributions/test/pom.xml
```diff
@@ -0,0 +1,135 @@
...
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <skip>true</skip>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>-->
```
kcover: I will remove this in the next commit.

Jenkinsfile
```diff
@@ -32,6 +32,8 @@ pipeline {
         LINUX_MVN_RANDOM = '-Djava.security.egd=file:/dev/./urandom'
         COVERAGE_EXCLUSIONS = '**/test/**/*,**/itests/**/*,**/*Test*,**/sdk/**/*,**/*.js,**/node_modules/**/*,**/jaxb/**/*,**/wsdl/**/*,**/nces/sws/**/*,**/*.adoc,**/*.txt,**/*.xml,/**/IT*.java'
         DISABLE_DOWNLOAD_PROGRESS_OPTS = '-Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn '
+        SOURCE_NODE = 'https://ditto-1.phx.connexta.com:8993/services'
```
clockard: I don't think we want these properties in our jenkins file. We should be able to pass them in through the settings file or maybe even just the jenkins job itself.

clockard: We might need to add a custom replication settings.xml that defines these properties and then include that settings in our withMaven call

paouelle: These should be defined on Jenkins and not checked in our repo at all. We should document that we require these properties to be defined. The goal is not to hardcode knowledge of the environment where we will be building.

pom.xml
```diff
@@ -312,6 +312,16 @@
     </dependencyManagement>
 
     <dependencies>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>${slf4j.version}</version>
```
paouelle: ✏️ These 2 dependencies should be duplicated in the `dependencyManagement` section along with the `<version>` and in here, just leave them but remove the `<version>`

distributions/test/README.md
```diff
@@ -0,0 +1,8 @@
+# Replication E2E Tests
+Running the Replication end to end tests requires a local docker instance to be running and the
```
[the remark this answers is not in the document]
kcover: By docker instance I meant docker itself. As in the docker daemon. I'll try to communicate that better.

PR #149, opened 2019-07-29
REP-148 Add metadata only replication option

### LINKED ISSUE #148: Add metadata replication only option
Add an option to replication only metadata. This has been requested several times. Replicated metadata should point back to original source of data for resource download.
Update:
Keeping the original resource download link caused an issue with metacard ids being regenerated. If a metacard is ingested with a resource uri not local to the current DDF a new ID is generated which causes problems correlating what has already been replicated. For now going to remove link back to original resource.

clockard:
#### What does this PR do?
Add metadata only replication option

Setup a replication job with the metadata only option checked. 
Verify metadata only replicated. This can be verified by checking that the resource-uri is empty/DNE.

Fixes #148 

Screen Shot 2019-07-29 at 1 47 01 PM

### DIFF
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
@@ -92,10 +92,14 @@ public abstract class AbstractPersistable implements Persistable {
    * before performing its own functionality.
    *
    * @param properties the properties to read into the persistable's variables
+   * @return int the version of the map passed in.
    */
-  public void fromMap(Map<String, Object> properties) {
+  public int fromMap(Map<String, Object> properties) {
     setId((String) properties.get(ID_KEY));
-    setVersion((int) properties.get(VERSION_KEY));
     setModified((Date) properties.get(MODIFIED_KEY));
+    if (!properties.containsKey(VERSION_KEY)) {
+      throw new IllegalStateException("Serialized properties map is expected to contain a version");
+    }
+    return (int) properties.get(VERSION_KEY);
   }
 }
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
@@ -132,15 +132,15 @@ public class ReplicationSiteImpl extends AbstractPersistable implements Replicat
   }
 
   @Override
-  public void fromMap(Map<String, Object> properties) {
-    super.fromMap(properties);
-    final int version = super.getVersion();
+  public int fromMap(Map<String, Object> properties) {
+    int serializedVersion = super.fromMap(properties);
 
-    if (version == ReplicationSiteImpl.CURRENT_VERSION) {
+    if (serializedVersion == ReplicationSiteImpl.CURRENT_VERSION) {
       fromCurrentMap(properties);
     } else {
-      fromIncompatibleMap(version, properties);
+      fromIncompatibleMap(serializedVersion, properties);
     }
+    return serializedVersion;
   }
 
   private void fromCurrentMap(Map<String, Object> properties) {
@@ -151,13 +151,11 @@ public class ReplicationSiteImpl extends AbstractPersistable implements Replicat
   }
 
   private void fromIncompatibleMap(int version, Map<String, Object> properties) {
-    switch (version) {
-      case 1:
-        fromVersionOneMap(properties);
-        break;
-      default:
-        LOGGER.error("unsupported {} version: {}", ReplicationSiteImpl.PERSISTENCE_TYPE, version);
-        throw new IllegalStateException("Unsupported version: " + version);
+    if (version == 1) {
+      fromVersionOneMap(properties);
+    } else {
+      LOGGER.error("unsupported {} version: {}", ReplicationSiteImpl.PERSISTENCE_TYPE, version);
+      throw new IllegalStateException("Unsupported version: " + version);
     }
     super.setVersion(CURRENT_VERSION);
   }
@@ -165,14 +163,14 @@ public class ReplicationSiteImpl extends AbstractPersistable implements Replicat
   private void fromVersionOneMap(Map<String, Object> properties) {
     setName((String) properties.get(NAME_KEY));
 
-    String oldUrl = (String) properties.get(URL_KEY);
+    String oldUrlStr = (String) properties.get(URL_KEY);
     try {
-      URL url = new URL(oldUrl);
-      if (StringUtils.isEmpty(url.getPath())) {
-        setUrl(oldUrl + "/services");
+      URL oldUrl = new URL(oldUrlStr);
+      if (StringUtils.isEmpty(oldUrl.getPath())) {
+        setUrl(oldUrlStr + "/services");
       } else {
         // should never hit this since old urls do not have context paths
-        setUrl(oldUrl);
+        setUrl(oldUrlStr);
       }
     } catch (MalformedURLException e) {
       // saved URLs will not have invalid URLs
--- a/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
+++ b/replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
@@ -107,8 +107,8 @@ public class ReplicatorConfigImpl extends AbstractPersistable implements Replica
   }
 
   @Override
-  public void fromMap(Map<String, Object> properties) {
-    super.fromMap(properties);
+  public int fromMap(Map<String, Object> properties) {
+    int serializedVersion = super.fromMap(properties);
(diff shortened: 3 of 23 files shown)


replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -113,6 +119,7 @@ public void fromMap(Map<String, Object> properties) {
     setSuspended(Boolean.valueOf((String) properties.get(SUSPENDED_KEY)));
     setDeleted(Boolean.valueOf((String) properties.get(DELETED_KEY)));
     setDeleteData(Boolean.valueOf((String) properties.get(DELETE_DATA_KEY)));
+    setMetadataOnly(Boolean.valueOf((String) properties.get(METADATA_ONLY_KEY)));
```
paouelle: ❗️ Looks to me like we should be checking the version we are getting and initializing the new properties with their default instead of blindly trying to extract values from a map. The current implementation can lead to differences between what is documented and what might happen

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -55,10 +57,11 @@
    *         <li>Add <b>suspended</b> field of type boolean with default of false
    *         <li>Add <b>deleted</b> field of type boolean with default of false
    *         <li>Add <b>deleteData</b> field of type boolean with default of false
+   *         <li>Add <b>metadataOnly</b> field of type boolean with default of false
```
peterhuffer: ✏️ This should go under a version 2 list.

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImplTest.java
```diff
@@ -65,7 +65,7 @@ public void writeToMap() {
     assertThat(props.get(FILTER_KEY), is("filter"));
```
peterhuffer: ❓ Should we add a test for reading from version 1 to version 2?

clockard: It was kind of already doing that but I added a more explicit test for it

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImplTest.java
```diff
@@ -90,5 +118,6 @@ public void readFromMap() {
     assertThat(config.isBidirectional(), is(true));
     assertThat(config.getFailureRetryCount(), is(5));
     assertThat(config.getVersion(), is(1));
```
peterhuffer: ❓ Shouldn't the version be 2 in this case?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
paouelle: These are assumptions (although most likely correct). I don't think we should rely on the presence or not of something to know which version it is; this is why we are saving a version number. We should rely on the version number and then not even look at the map when we know it shouldn't have been there to start with.

clockard: added a version check

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImplTest.java
clockard: I don't think so. The config is still version 1 and we are not converting it to a version 2

peterhuffer: Every time we read an old version though we are converting it to the latest version.

clockard: fixed

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/OldSite.java
```diff
@@ -40,5 +40,6 @@ public void fromMap(Map<String, Object> properties) {
 
     setVerifiedUrl(getUrl()); // do this after setUrl()
     setRemoteManaged(false);
+    return CURRENT_VERSION;
```
paouelle: I don't think this should return `CURRENT_VERSION`. It should return maybe `-1` since these sites had no version

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
```diff
@@ -132,15 +132,15 @@ public boolean isRemoteManaged() {
   }
 
   @Override
-  public void fromMap(Map<String, Object> properties) {
-    super.fromMap(properties);
-    final int version = super.getVersion();
+  public int fromMap(Map<String, Object> properties) {
+    int propVersion = super.fromMap(properties);
```
paouelle: ✏️ How about renaming it `currentVersion` or `serializedVersion`?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -92,10 +92,11 @@ private void setModified(@Nullable Date modified) {
...
    *
    * @param properties the properties to read into the persistable's variables
+   * @return int the version of the map passed in.
    */
-  public void fromMap(Map<String, Object> properties) {
+  public int fromMap(Map<String, Object> properties) {
     setId((String) properties.get(ID_KEY));
-    setVersion((int) properties.get(VERSION_KEY));
     setModified((Date) properties.get(MODIFIED_KEY));
+    return (int) properties.get(VERSION_KEY);
```
paouelle: ✏️ Just to be safe, shouldn't we check first if the property exist and throw an `IllegalStateException` if it doesn't instead of the NPE or `ClassCastException` that would be thrown by the current code?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -107,8 +107,8 @@ public ReplicatorConfigImpl() {
   }
 
   @Override
-  public void fromMap(Map<String, Object> properties) {
-    super.fromMap(properties);
+  public int fromMap(Map<String, Object> properties) {
+    int propVersion = super.fromMap(properties);
```
paouelle: ✏️ How about renaming it `currentVersion` or `serializedVersion`?

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -119,9 +119,10 @@ public void fromMap(Map<String, Object> properties) {
     setSuspended(Boolean.valueOf((String) properties.get(SUSPENDED_KEY)));
     setDeleted(Boolean.valueOf((String) properties.get(DELETED_KEY)));
     setDeleteData(Boolean.valueOf((String) properties.get(DELETE_DATA_KEY)));
-    if (getVersion() >= 2) {
+    if (propVersion >= 2) {
```
paouelle: ✏️ I prefer and approach like in `ReplicationSiteimpl` as it make sit easier to maintain and see what changed over time as opposed to performing the logic for handling different versions inside a single method.

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/AbstractPersistable.java
clockard: Added check

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicationSiteImpl.java
clockard: Changed to serializedVersion

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/data/ReplicatorConfigImpl.java
clockard: Changed to serializedVersion

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/legacy/OldSite.java
clockard: Changed to 0 since not having a version number was considered version 0

kcover:
Hero Successful ✔️

PR #161, opened 2019-08-01
Fix javadoc to work with latest plugin

clockard:
#### What does this PR do?
Fix javadoc to work with latest plugin

Manual javadoc build since CI on a PR will not run the javadoc

Updated the javadoc plugin via dependabot which resulted in stricter javadoc checking.
Some of the javadoc added for interfaces copied from DDF is not very descriptive but I think thats ok for now

### DIFF
--- a/replication-api/src/main/java/com/connexta/ion/replication/api/Replicator.java
+++ b/replication-api/src/main/java/com/connexta/ion/replication/api/Replicator.java
@@ -22,6 +22,7 @@ public interface Replicator {
    * Submits a {@link SyncRequest} to be executed
    *
    * @param syncRequest The {@link SyncRequest} to run
+   * @throws InterruptedException on thread interrupt when inserting request into the queue
    */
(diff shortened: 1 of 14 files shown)


replication-api/src/main/java/com/connexta/ion/replication/api/Replicator.java
```diff
@@ -22,6 +22,8 @@
    * Submits a {@link SyncRequest} to be executed
    *
    * @param syncRequest The {@link SyncRequest} to run
+   *
+   * @throws InterruptedException on error inserting request into the queue
```
paouelle: ❗️ This is not why `InterruptedException` would be thrown. It would be thrown because the thread was interrupted while trying to insert the request into the queue.

clockard: Updated

PR #180, opened 2019-08-09
Update replication to work with DDF 2.16.x

clockard:
#### What does this PR do?
Updates DDF version 
Still need to fix itest
Basic replication testing on a DDF 2.16.x
Will need to install the admin console 1.3.2 as well

### DIFF
--- a/distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
+++ b/distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
@@ -43,6 +43,7 @@ import org.codice.pax.exam.junit.ConfigurationAdmin;
 import org.codice.pax.exam.junit.ServiceAdmin;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.slf4j.Logger;
--- a/pom.xml
+++ b/pom.xml
@@ -26,9 +26,9 @@
         <snapshots.repository.url />
         <releases.repository.url />
 
-        <admin-query.version>1.2.8</admin-query.version>
+        <admin-query.version>1.3.2</admin-query.version>
         <ddf.support.version>2.3.12</ddf.support.version>
-        <ddf.version>2.13.10</ddf.version>
+        <ddf.version>2.16.1</ddf.version>
         <codice-maven.version>0.2</codice-maven.version>
         <codice-test.version>0.9</codice-test.version>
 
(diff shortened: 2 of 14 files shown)


distributions/replication-dominion-pax-exam/pom.xml
```diff
@@ -35,13 +35,13 @@
         <dependency>
             <groupId>ddf.lib</groupId>
             <artifactId>dominion-commons</artifactId>
-            <version>${ddf.version}</version>
+            <version>${dominion.version}</version>
```
paouelle: ❓ Is this because the dominion libraries have not been ported to the DDF branch we are now referencing? If so then this might actually cause a problem and these libraries should be ported over. Travis was working on that so we might want to check with him.

The fact is that the artifact version you are referencing is designed for the release branch and not the master branch of DDF which will cause issues or even not start properly DDF in some cases.

distributions/replication-dominion/src/main/filtered-resources/features.xml
```diff
@@ -20,9 +20,9 @@
     <repository>mvn:org.codice.test/dominion-pax-exam-feature/${codice-test.version}/xml/features
     </repository>
     <repository>mvn:ddf.lib/dominion-commons/${ddf.version}/xml/features</repository>
-    <repository>mvn:org.codice.ddf/ddf-dominion/${ddf.version}/xml/features</repository>
+    <repository>mvn:org.codice.ddf/ddf-dominion/${dominion.version}/xml/features</repository>
 
-    <feature name="ddf-dominion" version="${project.version}">
+    <feature name="dominion" version="${project.version}">
```
paouelle: ✏️ Not sure we should be renaming this one since it is the DDF extensions to dominion

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/QueryServiceImpl.java
```diff
@@ -61,42 +48,24 @@ public QueryServiceImpl(
    */
   @Override
   public Stream<Query> queries() {
-    Filter filter = builder.attribute(Core.METACARD_TAGS).is().equalTo().text("workspace");
+    Filter filter = builder.attribute(Core.METACARD_TAGS).is().equalTo().text("query");
     PropertyName sortProperty = new PropertyNameImpl(Core.METACARD_MODIFIED);
     SortBy sortPolicy = new SortByImpl(sortProperty, SortOrder.DESCENDING);
     QueryImpl query = new QueryImpl(filter, 1, 100, sortPolicy, false, 0L);
-    QueryResponse queryResponse;
+    QueryResponse queryTypeResponse;
```
kcover: queryTypeResponse?

distributions/replication-dominion/src/main/filtered-resources/features.xml
clockard: changed the name back to ddf-dominion

distributions/replication-dominion-pax-exam/pom.xml
clockard: Commented out all modules that use dominion and changed the versions back to just ddf.version

distributions/test/itests/src/test/java/org/codice/ditto/replication/admin/query/ITReplicationQuery.java
```diff
@@ -43,6 +43,7 @@
 import org.codice.pax.exam.junit.ServiceAdmin;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
```
peterhuffer: ❓ Unused?

pom.xml
```diff
@@ -26,9 +26,9 @@
         <snapshots.repository.url />
         <releases.repository.url />
 
-        <admin-query.version>1.2.8</admin-query.version>
+        <admin-query.version>1.3.2</admin-query.version>
         <ddf.support.version>2.3.12</ddf.support.version>
-        <ddf.version>2.13.10</ddf.version>
+        <ddf.version>2.16.1-SNAPSHOT</ddf.version>
```
peterhuffer: ❓ Should we just rely on the 2.16.0 version for now?

peterhuffer:
Successful hero. Verified creating a site, replication and running the replication. ✔️

PR #191, opened 2019-08-13
REPL-137 First version for the new queue broker API

### LINKED ISSUE #190: API support for queue Broker
Add API support for a Queue broker which will allow support for queuing up resources to replicate.

paouelle:
#### What does this PR do?
It defines interfaces for the new queue api which will be used by both query services and workers.

@clockard 
@kcover 
@peterhuffer 

PR build should suffice as no implementation or client exists yet.

Fixes #190 

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/replication-api/src/main/java/com/connexta/replication/queue/CompoundQueue.java
+++ b/replication-api/src/main/java/com/connexta/replication/queue/CompoundQueue.java
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.queue;
+
+import java.util.Optional;
+import java.util.stream.Stream;
+
+/**
+ * A compound queue is an artifact created around multiple {@link SiteQueue}s allowing a worker to
+ * retrieve tasks from any of the queues based on task priorities.
+ */
+public interface CompoundQueue extends Queue {
+  /**
+   * Gets all site queues that are compounded together.
+   *
+   * @return a stream of all site queues compounded together (never empty)
+   */
+  public Stream<SiteQueue> queues();
+
+  /**
+   * Gets all sites that have their queues compounded together.
+   *
+   * @return a stream of all site ids that have their queues compounded together (never empty)
+   */
+  public Stream<String> sites();
+
+  /**
+   * Gets a compounded queue for a given site.
+   *
+   * @param site the site for which to get a queue that was compounded
+   * @return the site queue to use for the specified site or empty if no queue was not compounded
+   *     for the specified site
+   */
+  public Optional<SiteQueue> getQueue(String site);
+}
--- a/replication-api/src/main/java/com/connexta/replication/queue/DataInfo.java
+++ b/replication-api/src/main/java/com/connexta/replication/queue/DataInfo.java
@@ -1,0 +1,34 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.queue;
+
+import java.util.OptionalLong;
+
+/** Identifies a particular piece of data (e.g. metadata, resource) to be transferred. */
+public interface DataInfo {
+  /**
+   * Gets the last modified timestamp from the source site for the piece of data that should be
+   * transferred by the task.
+   *
+   * @return last modified timestamp for the piece data
+   */
+  public long getLastModified();
+
+  /**
+   * Gets the size of the piece of data to be transferred if known.
+   *
+   * @return the optional size of the piece of data
+   */
+  public OptionalLong getSize();
+}
--- a/replication-api/src/main/java/com/connexta/replication/queue/DataInfo.java
+++ b/replication-api/src/main/java/com/connexta/replication/queue/DataInfo.java
@@ -13,6 +13,7 @@
  */
 package com.connexta.replication.queue;
 
+import java.time.Instant;
 import java.util.OptionalLong;
 
 /** Identifies a particular piece of data (e.g. metadata, resource) to be transferred. */
@@ -23,7 +24,7 @@ public interface DataInfo {
    *
    * @return last modified timestamp for the piece data
    */
-  public long getLastModified();
+  public Instant getLastModified();
 
   /**
    * Gets the size of the piece of data to be transferred if known.
--- a/replication-api/src/main/java/com/connexta/replication/queue/OperationType.java
+++ b/replication-api/src/main/java/com/connexta/replication/queue/OperationType.java
@@ -1,0 +1,39 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.queue;
+
+/**
+ * Defines the various type of operations a worker can perform with an intel document for the site
+ * associated with the queue where a task has been added.
+ */
+public enum OperationType {
+  /** Harvests a given intel document from the site. */
+  HARVEST,
+
+  /** Replicates data for a given intel document to the site. */
+  REPLICATE_TO,
+
+  /** Replicates data for a given intel document from the site. */
+  REPLICATE_FROM,
+
+  /** Deletes a given intel document from the site. */
+  DELETE,
+
+  /**
+   * The unknown value is used for forward compatibility where a worker might not be able to
+   * understand a new type of operation and would mapped this new operation to <code>UNKNOWN</code>
+   * and most likely ignore it without removing it from the queue.
+   */
+  UNKNOWN
+}
--- a/replication-api/src/main/java/com/connexta/replication/queue/PriorityLevel.java
+++ b/replication-api/src/main/java/com/connexta/replication/queue/PriorityLevel.java
@@ -1,0 +1,42 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.queue;
+
+/** Defines the various priority levels supported for queued tasks. */
+public enum PriorityLevel {
+  HIGH((byte) 9),
+  MEDIUM((byte) 5),
+  LOW((byte) 0),
+
+  /**
+   * The unknown value is used for forward compatibility where a worker might not be able to
+   * understand this new priority and would mapped it automatically to priority level 1.
+   */
+  UNKNOWN((byte) 1);
+
+  private final byte level;
+
+  private PriorityLevel(byte level) {
+    this.level = level;
+  }
+
+  /**
+   * Gets the level for this priority ranging from 9 being the highest level to 0 being the lowest.
+   *
+   * @return the level for this priority
+   */
+  public byte getLevel() {
+    return level;
+  }
+}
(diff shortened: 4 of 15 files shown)


replication-api/src/main/java/com/connexta/replication/queue/ResourceInfo.java
```diff
@@ -0,0 +1,26 @@
...
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.queue;
+
+import java.net.URI;
+import java.util.Optional;
+
+public interface ResourceInfo extends DataInfo {
+  /**
+   * Gets the URI for the resource or empty if there is none ir if it is not known.
```
peterhuffer: ✏️ `... or empty if it is now known.`

replication-api/src/main/java/com/connexta/replication/queue/ResourceInfo.java
```diff
@@ -0,0 +1,26 @@
...
+   *
+   * @return the optional URI for the resource
+   */
+  public Optional<URI> getResourceUri();
```
peterhuffer: ❓ In which case would the uri of a resource be unknown?

replication-api/src/main/java/com/connexta/replication/queue/QueueBroker.java
```diff
@@ -0,0 +1,105 @@
...
+
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
+import javax.annotation.Nullable;
+
+/**
+ * A queue broker defines the point of entry for retrieving references to queues and for monitoring
+ * multiple queues at the same time for tasks to be processed.
+ */
+public interface QueueBroker {
```
peterhuffer: We will need to expose statistics about queues so we can collect metrics on them. Doesn't need to be done as part of this PR though.

replication-api/src/main/java/com/connexta/replication/queue/Task.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Represents a task that was queued for later processing by a worker.
+ *
+ * <p>When a task is first queued, its total attempts counter will report <code>0</code> and its
+ * queued time and original queued time will be set to the same time (i.e. the current time when the
+ * task is queued via one of {@link Queue#offer(TaskInfo)}), {@link Queue#offer(TaskInfo, long,
+ * TimeUnit)}, or {@link Queue#put(TaskInfo)}. If the task fails and a decision is made to put it
+ * back into the queue via one of {@link #offerForRetry(PriorityLevel)}, {@link
```
peterhuffer: ❓ Doesn't a Task only get removed from the queue when it is successfully performed? In which case we would never put the task back on the queue?

replication-api/src/main/java/com/connexta/replication/queue/Task.java
```diff
@@ -0,0 +1,154 @@
...
+   * #getOriginalQueuedTime()} if the task has not yet been attempted ({@link #getTotalAttempts()}
+   * returns <code>0</code>).
+   *
+   * @return the time when this task was queued in milliseconds from the epoch
+   */
+  public long getQueuedTime();
+
+  /**
+   * Gets the total number of times this particular task has been attempted. When a task fails for a
+   * reason that doesn't preclude it from being re-tried and it gets re-queued by the worker, its
```
peterhuffer: This re-queuing is confusing me. What reason would a worker choose to evict a Task from the queue other than a success and why does a worker get to make that decision?

replication-api/src/main/java/com/connexta/replication/queue/Task.java
```diff
@@ -0,0 +1,154 @@
...
+   */
+  public int getTotalAttempts();
+
+  /**
+   * Checks if this task is still locked by the current thread/worker. A task will automatically be
+   * unlocked once one of {@link #unlock()}, {@link #complete()}, {@link #fail()}, {@link
+   * #offerForRetry(PriorityLevel)}, {@link #offerForRetry(PriorityLevel, long, TimeUnit)}, or
+   * {@link #putForRetry(PriorityLevel)} has been called on it.
+   *
+   * @return <code>true</code> if the current thread/worker owns the lock to this task; <code>false
```
peterhuffer: ❓ Why the `current` thread/worker? I figured this would be used by the broker, when a Task is requested, to skip locked Tasks that haven't yet been removed from the queue.

replication-api/src/main/java/com/connexta/replication/queue/Task.java
```diff
@@ -0,0 +1,154 @@
...
+   * Reports the failed completion of this task. The task will automatically be unlocked and removed
+   * from the associated queue. Such a task will not be re-queued for another attempt at being
+   * processed.
+   *
+   * @throws IllegalStateException if the current thread/worker doesn't own the lock to the task or
+   *     if the task is not currently or no longer locked
+   * @throws InterruptedException if the thread was interrupted while attempting to unlock the task
+   * @throws QueueException if an error occurred while performing the operation
+   */
+  public void fail() throws QueueException, InterruptedException;
```
peterhuffer: ❓ Would it be worth passing a reason string for better reporting experience? May not be needed right away.

replication-api/src/main/java/com/connexta/replication/queue/Task.java
```diff
@@ -0,0 +1,154 @@
...
+   *     </code> otherwise
+   */
+  public boolean isLocked();
+
+  /**
+   * Attempts to unlock this task. An unlock task will be picked up by the next worker attempting to
+   * poll for a task from the associated queue.
+   *
+   * @throws IllegalStateException if the current thread/worker doesn't own the lock to the task or
+   *     if the task is not current locked
```
peterhuffer: ✏️ `current` -> `currently`

replication-api/src/main/java/com/connexta/replication/queue/Task.java
```diff
@@ -0,0 +1,154 @@
...
+   * after a failure occurs which permits another attempt at processing the task.
+   *
+   * @param level the new level for the new re-queued task
+   * @return <code>true</code> if the task was removed and re-added to this queue (only then will
+   *     the task be unlocked); <code>false</code> otherwise (the task is still considered locked)
+   * @throws IllegalStateException if the current thread/worker doesn't own the lock to the task or
+   *     if the task is not currently or no longer locked
+   * @throws QueueException if an error occurred while performing the operation
+   */
+  public boolean offerForRetry(PriorityLevel level) throws QueueException;
```
peterhuffer: ❓ It seems like with the way this works now, a Task can be offered back into the queue with the same priority, but it will lose its original position, whereas if we only removed on success, a failure would keep its original position. Is losing the original position acceptable?
❓ Should a Worker be allowed to re-queue a Task at any priority? What happens when a priority service exists later and the priority it sets is violated by a worker?

replication-api/src/main/java/com/connexta/replication/queue/TaskInfo.java
```diff
@@ -0,0 +1,66 @@
...
+   *
+   * @return the info for the resource that should be transferred or empty if none should be
+   *     transferred
+   */
+  public Optional<ResourceInfo> getResource();
+
+  /**
+   * Gets the information for the different types of metadata that should be transferred.
+   *
+   * @return all the different types of metadata that should be transferred or empty if none should
```
peterhuffer: ❓ Can this be empty? Would we ever transfer a resource without corresponding metadata?

replication-api/src/main/java/com/connexta/replication/queue/OperationType.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+package com.connexta.replication.queue;
+
+/**
+ * Defines the various type of operations a worker can perform with an intel document for the site
+ * associated with the queue where a task has been added.
+ */
+public enum OperationType {
+  /** Harvests a given intel document from the site. */
+  HARVEST,
```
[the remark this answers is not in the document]
paouelle: Discussed offline.

replication-api/src/main/java/com/connexta/replication/queue/PriorityLevel.java
```diff
@@ -0,0 +1,42 @@
...
+  UNKNOWN((byte) 1);
+
+  private final byte level;
+
+  private PriorityLevel(byte level) {
+    this.level = level;
+  }
+
+  /**
+   * Gets the level for this priority ranging from 9 being the highest level to 0 being the lowest.
```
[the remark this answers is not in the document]
paouelle: Typical in queue providers such as ActiveMQ and it leaves us some flexibility in the future to go beyond just low, medium, high if we ever need to.

replication-api/src/main/java/com/connexta/replication/queue/Queue.java
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * A queue defines all possible operations that can be performed. Queues are specific to a given
+ * site and will hold tasks which will specify intel that either needs to be harvested from,
+ * retrieved from, or replicated to. For example:
+ *
+ * <ul>
+ *   <li>a Tactical Ion or DIB site we replicate from/to
+ *   <li>a DDF-based site we are harvesting from or replicating to
+ *   <li>a Regional Ion site to which we are replicating our local intel to
+ *   <li>the local Ion site (typically will be used for delete operations only)
```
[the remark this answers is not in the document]
paouelle: Yes. However, I am trying very hard not to use the term `node` since a site will be composed of multiple nodes since we are cloud based. So the local site is the site we are deployed to be part of.

replication-api/src/main/java/com/connexta/replication/queue/Queue.java
```diff
@@ -0,0 +1,107 @@
...
+  /**
+   * Retrieves the next available highest priority task, waiting if necessary until one becomes
+   * available. The returned task is considered locked by the current thread/worker.
+   *
+   * @return the next available task with the highest priority from this queue
+   * @throws InterruptedException if the current thread was interrupted while waiting for a task to
+   *     be returned
+   * @throws QueueException if an error occurred while performing the operation
+   */
+  public Task take() throws QueueException, InterruptedException;
```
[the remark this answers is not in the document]
paouelle: Typical queue interfaces makes the distinction because taking means you are willing to wait forever for a task while polling means you are willing to wait for a maximum amount of time before control should be returned to you with no tasks.

replication-api/src/main/java/com/connexta/replication/queue/Task.java
paouelle: This is part of the API used by clients. The state you are talking about is not something a client would need or care about and as such can be left hidden away as part of the broker's implementation of this interface.

replication-api/src/main/java/com/connexta/replication/queue/Task.java
paouelle: These were changed after an offline discussions.

replication-api/src/main/java/com/connexta/replication/queue/Task.java
paouelle: These were changed after an offline discussions.

replication-api/src/main/java/com/connexta/replication/queue/TaskInfo.java
paouelle: Maybe. For Ion to Ion (non tactical) we might just transfer the resource and let the other side regenerate the metadata. The current API allows us the freedom to support both when we have a need to make that decision.

replication-api/src/main/java/com/connexta/replication/queue/CompoundQueue.java
```diff
@@ -0,0 +1,46 @@
...
+package com.connexta.replication.queue;
+
+import java.util.Optional;
+import java.util.stream.Stream;
+
+/**
+ * A compound queue is an artifact created around multiple {@link SiteQueue}s allowing a worker to
+ * retrieve tasks from any of the queues based on task priorities.
+ */
+public interface CompoundQueue extends Queue {
```
peterhuffer: ✏️ Consider `CompositeQueue`?

peterhuffer: I don't think we should spend much churn over this, but this may be able to just be an implementation that is returned via the broker and a client doesn't care which queues are part of the queue. If a client needs the SiteQueue to offer/put to, then it can go individually through the broker?

replication-api/src/main/java/com/connexta/replication/queue/Task.java
peterhuffer: Why does a client care if a Task it has is locked. By virtue of the broker giving a client the Task, it should be locked for that client, right?

replication-api/src/main/java/com/connexta/replication/queue/CompoundQueue.java
```diff
@@ -0,0 +1,46 @@
...
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.queue;
+
+import java.util.Optional;
+import java.util.stream.Stream;
+
+/**
+ * A compound queue is an artifact created around multiple {@link SiteQueue}s allowing a worker to
```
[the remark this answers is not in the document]
paouelle: 🤔 I would suspect each worker would have its own queue instance so I think `worker` is still the right thing.

replication-api/src/main/java/com/connexta/replication/queue/CompoundQueue.java
paouelle: Sure. Sugar with that :-D

paouelle: I don't think there will be any churn on this since this composite queue will not to keep track of the set of `SiteQueue`s anyway. So this is a strait and simple getter.

replication-api/src/main/java/com/connexta/replication/queue/CompoundQueue.java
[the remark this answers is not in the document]
peterhuffer: Whoops, misread that nvm.

replication-api/src/main/java/com/connexta/replication/queue/Task.java
paouelle: A client might not care if properly implemented. However, a client which cannot imply the state based on its implementation shouldn't blindly call the methods below and simply catch the `IllegalStateException`. The contract of the methods below indicates you should not be calling them if you are not the owner so it becomes the responsibility of the client to ensure either by its design or manually by checking that they are the owner of a task before calling these methods.

In addition, implementation of this API will end up first checking if the caller if the owner of the task (or lock) by calling this exact method.

In the end, I am not exposing something they don't know about. The lock concept is part of this API and a client has to be aware of it.

replication-api/src/main/java/com/connexta/replication/queue/Task.java
```diff
@@ -0,0 +1,128 @@
...
+   * unlocked once one of {@link #unlock()}, {@link #complete()}, {@link #fail(ErrorCode)}, or
+   * {@link #fail(ErrorCode, String)} has been called on it.
+   *
+   * @return <code>true</code> if the current thread/worker owns the lock to this task; <code>false
+   *     </code> otherwise
+   */
+  public boolean isLocked();
+
+  /**
+   * Attempts to unlock this task. An unlock task will be picked up by the next worker attempting to
```
kcover: ✏️ unlocked*

replication-api/src/main/java/com/connexta/replication/queue/DataInfo.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.OptionalLong;
+
+/** Identifies a particular piece of data (e.g. metadata, resource) to be transferred. */
+public interface DataInfo {
+  /**
+   * Gets the last modified timestamp from the source site for the piece of data that should be
+   * transferred by the task.
+   *
+   * @return last modified timestamp for the piece data
```
peterhuffer: ❓ Is this in milliseconds?

paouelle: Good catch. I am thinking of changing it to `Instant` now (same for all others).

PR #196, opened 2019-08-14
Add filter index status

peterhuffer:
#### What does this PR do?
- Add a separate Filter Index to track what should be queried from a remote site.
- Reduce logging in integration tests

@paouelle @kcover @clockard 

CI

This index is taken out of the current ReplicatorConfig which currently tracks the last modified date (modified since). It will be used by the upcoming query service which populates the queue of tasks.

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests


replication-api/src/main/java/com/connexta/ion/replication/api/data/Persistable.java
```diff
@@ -23,13 +23,6 @@
    */
   String getId();
 
-  /**
```
paouelle: ❗️ See above comment

replication-api-impl/src/main/java/com/connexta/ion/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -40,11 +40,6 @@ public String getId() {
     return id;
   }
 
-  @Override
```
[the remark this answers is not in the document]
peterhuffer: If we keep it part of the interface it makes the ID of each of these mutable. If a setter for the ID is needed, it can be on the pojo.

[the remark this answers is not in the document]
paouelle: But by removing it, you have now broken all other persistable objects right?

replication-api/src/main/java/com/connexta/ion/replication/api/data/SiteIndex.java
```diff
@@ -0,0 +1,31 @@
...
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.ion.replication.api.data;
+
+import java.util.Date;
+
+/** Keeps track of what to query from a {@link ReplicationSite}. */
```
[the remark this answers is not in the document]
peterhuffer: I figured anything would go into this index that will help determine what to query from a site and that the name was not limiting to what could go into this configuration.

replication-api-impl/src/main/java/com/connexta/ion/replication/api/impl/data/AbstractPersistable.java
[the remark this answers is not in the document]
peterhuffer: Ya but they were fixed as part of this PR.

replication-api/src/main/java/com/connexta/ion/replication/api/data/SiteIndex.java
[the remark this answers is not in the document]
paouelle: ok. So how about something like `SiteQueryIndex`?

replication-api-impl/src/main/java/com/connexta/ion/replication/api/impl/data/SiteIndexImpl.java
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Creates a new SiteIndex.
+   *
+   * @param id the unique id of this index
+   * @param version the version of this index
+   * @param modifiedSince the date to query resources that have been modified since
+   * @param siteId the site id this index applies to
+   */
+  public SiteIndexImpl(String id, int version, Date modifiedSince, String siteId) {
```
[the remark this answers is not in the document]
peterhuffer: Yeah it will use the all args ctor, but I cannot have the other ctor actually for solr to be able to recognize the all args without telling solr which ctor to use. So I may change the first ctor to a factory method. 
Solr will match the property name to the ctor name.
This implementation should be enforcing the interface contract right? And then the pojo + manager will take care of making sure we convert and detect objects correctly.

replication-api-impl/src/main/java/com/connexta/ion/replication/api/impl/data/SiteIndexImpl.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.Date;
+import java.util.UUID;
+import org.apache.commons.lang3.Validate;
+import org.springframework.data.annotation.Id;
+import org.springframework.data.solr.core.mapping.Indexed;
+import org.springframework.data.solr.core.mapping.SolrDocument;
+
+/** Simple implementation of SiteIndex. */
+@SolrDocument(collection = "site_index")
+public class SiteIndexImpl implements SiteIndex {
```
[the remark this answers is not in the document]
peterhuffer: I didn't really see a need for it. No reason in particular though.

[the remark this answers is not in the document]
paouelle: It already provided support for the id and the version and later on, it will end up redefining the equivalent `fromMap()` and `toMap()` for conversion.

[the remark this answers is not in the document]
peterhuffer: Won't that equivalent functionality be in the managers?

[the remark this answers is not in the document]
peterhuffer: I extended it for now.

replication-api-impl/src/main/java/com/connexta/ion/replication/api/impl/data/SiteQueryIndexImpl.java
```diff
@@ -0,0 +1,87 @@
...
+@SolrDocument(collection = "site_query_index")
+public class SiteQueryIndexImpl extends AbstractPersistable implements SiteQueryIndex {
+
+  private static final int CURRENT_VERSION = 1;
+
+  @Indexed(name = "modified_since")
+  private final Instant modifiedSince;
+
+  @Indexed(name = "site_id")
+  private final String siteId;
```
kcover: Should we be indexing by filter, rather than by site?

replication-api-impl/src/main/java/com/connexta/ion/replication/api/impl/persistence/SiteQueryIndexManagerImpl.java
```diff
@@ -0,0 +1,76 @@
...
+   * Creates a new SiteQueryIndexManager.
+   *
+   * @param sites repository for site indices
+   */
+  public SiteQueryIndexManagerImpl(SiteQueryIndexRepository sites) {
+    this.sites = sites;
+  }
+
+  @Override
+  public SiteQueryIndex getBySite(String site) {
```
kcover: ✏️ getByFilterID* so the parameter is obvious

replication-api-impl/src/main/java/com/connexta/ion/replication/api/impl/data/SiteQueryIndexImpl.java
paouelle: ☝️

replication-api-impl/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
paouelle: Thank you
❓ Should we add the copyright header?

replication-api/src/main/java/com/connexta/replication/api/data/FilterIndex.java
```diff
@@ -0,0 +1,32 @@
...
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.api.data;
+
+import com.connexta.ion.replication.api.data.Metadata;
+import java.time.Instant;
+
+/** Keeps track of what to query from a {@link com.connexta.replication.api.data.Site}. */
```
paouelle: ✏️ ... from a filter?

replication-api/src/main/java/com/connexta/replication/api/data/FilterIndex.java
```diff
@@ -0,0 +1,32 @@
...
+public interface FilterIndex extends Persistable {
+
+  /**
+   * A time which represents the modified date of the last {@link Metadata} that was put to the
+   * Queue.
+   *
+   * @return the {@link Instant}, or {@code null} if no metadata has attempted to be replicated
+   */
+  Instant getModifiedSince();
```
paouelle: ❗️ ❓ Why would such an object exist if we haven't tried to replicate yet? I would make it a required field and not support it being `null`

replication-api/src/main/java/com/connexta/replication/api/data/FilterIndex.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /** @return the site id this index is for */
```
paouelle: ✏️ Got to put more than the `@return`

replication-api/src/main/java/com/connexta/replication/api/data/FilterIndex.java
```diff
@@ -0,0 +1,32 @@
...
+package com.connexta.replication.api.data;
+
+import com.connexta.ion.replication.api.data.Metadata;
+import java.time.Instant;
+
+/** Keeps track of what to query from a {@link com.connexta.replication.api.data.Site}. */
+public interface FilterIndex extends Persistable {
+
+  /**
+   * A time which represents the modified date of the last {@link Metadata} that was put to the
```
paouelle: ✏️ I would avoid referencing the `Metadata` class here as it might change and for Ion, those are pieces of intel and not metadata.

replication-api/src/main/java/com/connexta/replication/api/persistence/DataManager.java
```diff
@@ -22,13 +22,6 @@
  */
 public interface DataManager<T extends Persistable> {
 
-  /**
```
paouelle: ❗️ ❓ Why are we removing this one?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * List of possible versions:
+   *
+   * <ul>
+   *   <li>1 - initial version.
+   * </ul>
+   */
+  public static final int CURRENT_VERSION = 1;
+
```
paouelle: ❗️ I am adding the following constant such that in the `readFrom()`, we can avoid hardcoding 0.

```
  /** The oldest version supported by the current code (anything before that will fail). */
  public static final int MINIMUM_VERSION = 1;
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
```diff
@@ -0,0 +1,110 @@
...
+  public static final String COLLECTION = "filter_index";
+
+  @Nullable
+  @Indexed(name = "modified_since", searchable = false)
+  private Instant modifiedSince;
+
+  @Nullable
```
paouelle: ✏️ nit - `filterId` should come before `modifiedSince` since it is part of what uniquely identifies this index

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
```diff
@@ -0,0 +1,110 @@
...
+  @Indexed(name = "filter_id")
```
paouelle: ❓ Should `required` be set to `true`?

solr-schema/src/main/resources/solr/conf/replication-solr-managed-schema.xml
```diff
@@ -176,6 +176,12 @@
     <field name="last_metadata_modified" type="pdate" indexed="true" stored="false" required="false"
            multiValued="false" />
 
+    <!-- site_index fields -->
+    <field name="modified_since" type="pdate" indexed="false" stored="true" required="false"
+      multiValued="false"/>
+    <field name="filter_id" type="string" indexed="true" stored="true" required="false"
```
paouelle: ❓ Should `required` be set to `true`?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
```diff
@@ -0,0 +1,110 @@
...
+  @Nullable
+  public Instant getModifiedSince() {
+    return modifiedSince;
+  }
+
+  /**
+   * @param modifiedSince the modified since time
+   * @return this pojo
+   */
+  public FilterIndexPojo setModifiedSince(Instant modifiedSince) {
```
paouelle: ✏️ got to annotate the parameter with `@Nullable`

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
```diff
@@ -0,0 +1,110 @@
...
+  @Nullable
+  public String getFilterId() {
+    return filterId;
+  }
+
+  /**
+   * @param filterId filter id for which this index belongs to
+   * @return this pojo
+   */
+  public FilterIndexPojo setFilterId(String filterId) {
```
paouelle: ✏️ got to annotate the parameter with `@Nullable`

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
```diff
@@ -0,0 +1,110 @@
...
+    this.filterId = filterId;
+    return this;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(super.hashCode(), modifiedSince, getFilterId());
```
paouelle:
```suggestion
    return Objects.hash(super.hashCode(), modifiedSince, filterId);
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
```diff
@@ -0,0 +1,110 @@
...
+          && Objects.equals(filterId, pojo.filterId);
+    }
+    return false;
+  }
+
+  @Override
+  public String toString() {
+    return String.format(
+        "FilterIndexPojo[id=%s, version=%d, modifiedSince=%s, filterId=%s]",
+        getId(), getVersion(), modifiedSince, filterId);
```
paouelle: ✏️ nit - I would put `filterId` before `modifiedSince`

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojoTest.java
```diff
@@ -0,0 +1,118 @@
...
+  private static final String FILTER_ID = "filterId";
+
+  private static final FilterIndexPojo POJO =
+      new FilterIndexPojo().setId(ID).setModifiedSince(MODIFIED_SINCE).setFilterId(FILTER_ID);
+
+  private FilterIndexPojo pojo;
+
+  @Before
+  public void setup() {
+    pojo = new FilterIndexPojo();
```
paouelle: ✏️ Doesn't need a setup section for that; simply do the new above where it is defined

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojoTest.java
```diff
@@ -0,0 +1,118 @@
...
+    final FilterIndexPojo pojo2 =
+        new FilterIndexPojo().setId(ID).setModifiedSince(MODIFIED_SINCE).setFilterId(FILTER_ID);
+    assertThat(POJO.equals(pojo2), is(true));
+    assertThat(POJO.equals(pojo2), is(true));
+  }
+
+  @Test
+  public void testEqualsNull() {
+    assertThat(POJO.equals(null), is(false));
+  }
```
paouelle: ❗️ Missing test cases that verify when passing a different class object and also when the various fields are different.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterIndexImpl.java
```diff
@@ -0,0 +1,99 @@
...
+  public String toString() {
+    return String.format(
+        "FilterIndexImpl[id=%s, modifiedSince=%s, filterId=%s]", getId(), modifiedSince, filterId);
+  }
+
+  @Override
+  public FilterIndexPojo writeTo(FilterIndexPojo pojo) {
+    super.writeTo(pojo);
+    setOrFailIfNullOrEmpty("filterId", this::getFilterId, pojo::setFilterId);
+    return pojo.setVersion(FilterIndexPojo.CURRENT_VERSION).setModifiedSince(modifiedSince);
```
paouelle: ❗️ The modified since field should also not be null

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterIndexImpl.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  @Override
+  public void readFrom(FilterIndexPojo pojo) {
+    super.readFrom(pojo);
+    if (pojo.getVersion() < FilterIndexPojo.CURRENT_VERSION) {
```
paouelle: ✏️ Change to `MINIMUM_VERSION` (see other comment on pojo)

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterIndexImpl.java
```diff
@@ -0,0 +1,99 @@
...
+      throw new ReplicationPersistenceException(
+          "unsupported "
+              + FilterIndexImpl.TYPE
+              + " version: "
+              + pojo.getVersion()
+              + " for object: "
+              + getId());
+    } // do support pojo.getVersion() > CURRENT_VERSION for forward compatibility
+    setOrFailIfNullOrEmpty("filterId", pojo::getFilterId, this::setFilterId);
```
paouelle: ✏️ I have move the rest into a new private method named `readFromCurrentAndFutureVersion` to prepare for upcoming version bumps

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterIndexImpl.java
```diff
@@ -0,0 +1,99 @@
...
+    setModifiedSince(pojo.getModifiedSince());
```
paouelle: ❗️ Should also not be null

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+  private static final Instant MODIFIED_SINCE = Instant.now();
+
+  private static final String FILTER_ID = "filterId";
+
+  private static final String ID = "id";
+
+  @Rule public ExpectedException exception = ExpectedException.none();
+
+  @Test
```
paouelle: ❗️ Should test the state of the default ctor

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/spring/FilterIndexRepository.java
```diff
@@ -0,0 +1,23 @@
...
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.api.impl.persistence.spring;
+
+import com.connexta.replication.api.impl.persistence.pojo.FilterIndexPojo;
+import java.util.Optional;
+import org.springframework.data.repository.CrudRepository;
+
+public interface FilterIndexRepository extends CrudRepository<FilterIndexPojo, String> {
```
clockard: I will let Patrick add a comment about javadoc here :)

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    FilterIndexImpl index = new FilterIndexImpl(MODIFIED_SINCE, FILTER_ID);
+    FilterIndexPojo pojo = index.writeTo(new FilterIndexPojo());
+    assertThat(pojo.getId(), notNullValue());
+    assertThat(pojo.getFilterId(), is(FILTER_ID));
+    assertThat(pojo.getModifiedSince(), is(MODIFIED_SINCE));
+    assertThat(pojo.getVersion(), is(FilterIndexPojo.CURRENT_VERSION));
+  }
+
+  @Test
+  public void testPojoCtorWritesTo() {
```
paouelle: ❓ `WritesTo`?

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+  private static final Instant MODIFIED_SINCE = Instant.now();
+
+  private static final String FILTER_ID = "filterId";
+
+  private static final String ID = "id";
+
+  @Rule public ExpectedException exception = ExpectedException.none();
+
+  @Test
+  public void testWriteTo() {
```
paouelle: ❗️ Should test `writeTo()` with each invalid fields as verified by the method

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    FilterIndexImpl index = new FilterIndexImpl();
+    index.readFrom(pojo);
+    assertThat(index.getFilterId(), is(FILTER_ID));
+    assertThat(index.getId(), is(ID));
+    assertThat(index.getModifiedSince(), is(MODIFIED_SINCE));
+  }
+
+  @Test
+  public void testReadFromFilterIdEmpty() {
+    exception.expect(ReplicationPersistenceException.class);
```
paouelle: ❓ How do you know it failed because of that?

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    FilterIndexPojo pojo =
+        new FilterIndexPojo().setFilterId("").setModifiedSince(MODIFIED_SINCE).setId(ID);
+    FilterIndexImpl index = new FilterIndexImpl();
+    index.readFrom(pojo);
+  }
+
+  @Test
+  public void testReadFromFilterIdNull() {
+    exception.expect(ReplicationPersistenceException.class);
```
paouelle: ❓ How do you know it failed because of that?

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+  @Test
+  public void testReadFromPreviousVersion() {
+    exception.expect(ReplicationPersistenceException.class);
+    exception.expectMessage("unsupported");
+    FilterIndexPojo pojo =
+        new FilterIndexPojo().setId(ID).setVersion(FilterIndexPojo.CURRENT_VERSION - 1);
+    new FilterIndexImpl(pojo);
+  }
+}
```
paouelle: ❗️ Should have one `testReadFromFutureVersion` to test forward compatibility

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    FilterIndexImpl index = new FilterIndexImpl();
+    index.readFrom(pojo);
+  }
+
+  @Test
+  public void testReadFromPreviousVersion() {
+    exception.expect(ReplicationPersistenceException.class);
+    exception.expectMessage("unsupported");
+    FilterIndexPojo pojo =
+        new FilterIndexPojo().setId(ID).setVersion(FilterIndexPojo.CURRENT_VERSION - 1);
```
paouelle: ✏️ Use the new MINIMUM_VERSION constant

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexImplTest.java
```diff
@@ -0,0 +1,94 @@
...
+    FilterIndexPojo pojo =
+        new FilterIndexPojo().setFilterId(FILTER_ID).setModifiedSince(MODIFIED_SINCE).setId(ID);
+    FilterIndexImpl index = new FilterIndexImpl();
+    index.readFrom(pojo);
+    assertThat(index.getFilterId(), is(FILTER_ID));
+    assertThat(index.getId(), is(ID));
+    assertThat(index.getModifiedSince(), is(MODIFIED_SINCE));
+  }
+
+  @Test
```
paouelle: ✏️ Should have tests for the modified since being null

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexManagerImplTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Test(expected = NotFoundException.class)
+  public void testGetBySiteNotFound() {
+    final String id = "id";
+    when(repository.findById(id)).thenReturn(Optional.empty());
+    sites.get(id);
+  }
+
+  @Test(expected = NotFoundException.class)
+  public void testGetBySiteIdNotFound() {
```
paouelle: ✏️ GetByFilter?

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexManagerImplTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+  @Mock FilterIndexRepository repository;
+
+  @Before
+  public void setup() {
+    sites = new FilterIndexManagerImpl(repository);
+  }
+
+  @Test(expected = NotFoundException.class)
+  public void testGetBySiteNotFound() {
```
paouelle: ❗️ Test when one is found, when a pojo is invalid

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexManagerImplTest.java
paouelle: ❗️ Test when one is found, when a pojo is invalid

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexManagerImplTest.java
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  @Test(expected = NotFoundException.class)
+  public void testGetBySiteIdNotFound() {
+    final String id = "id";
+    when(repository.findByFilterId(id)).thenReturn(Optional.empty());
+    sites.getByFilter(id);
+  }
+
+  @Test(expected = ReplicationPersistenceException.class)
```
paouelle: ❗️ Test with a valid object, with an invalid one

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexManagerImplTest.java
```diff
@@ -0,0 +1,61 @@
...
+  public void testSaveWrongObjectType() {
+    sites.save(mock(FilterIndex.class));
+  }
```
paouelle: ❗️ Need to test the other methods: `objects()`, `create()`, `remove()`

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/spring/FilterIndexRepository.java
```diff
@@ -0,0 +1,23 @@
...
+
+  Optional<FilterIndexPojo> findByFilterId(String filterId);
```
paouelle: javadoc

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/ReplicationItemManagerImpl.java
```diff
@@ -59,11 +59,6 @@ public ReplicationItemManagerImpl(ItemRepository itemRepository, SolrTemplate so
     this.solrTemplate = solrTemplate;
   }
 
-  @Override
```
paouelle: ❓ Same question about removing this one. Especially since the query service will need to create this object and should not create the impl directly

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/ReplicationItemManagerImpl.java
```diff
@@ -149,7 +144,7 @@ public void remove(String id) {
         new GroupOptions().addGroupByField("metadata_id").setLimit(1).addSort(doneTimeDescSort);
```
paouelle: ❓ Can there ever be more than one? Should it be ensured by the persistence layer that it never happens? Maybe by making it the unique id?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/ReplicatorConfigManagerImpl.java
```diff
@@ -29,11 +29,6 @@ public ReplicatorConfigManagerImpl(ConfigRepository configRepository) {
     this.configRepository = configRepository;
   }
 
-  @Override
```
paouelle: ❗️ Same question as before

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/SiteManagerImpl.java
```diff
@@ -29,11 +29,6 @@ public SiteManagerImpl(SiteRepository siteRepository) {
     this.siteRepository = siteRepository;
   }
 
-  @Override
```
paouelle: ❗️ Same question as before

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
```diff
@@ -0,0 +1,110 @@
...
+   * @return this pojo
+   */
+  public FilterIndexPojo setModifiedSince(Instant modifiedSince) {
+    this.modifiedSince = modifiedSince;
+    return this;
+  }
+
+  /** @return the filter id for this this index belongs to */
+  @Nullable
+  public String getFilterId() {
```
paouelle: ❓ Did we consider making this the actual id for the index as opposed to a field? This would ensure that only one object would ever exist for a given filter. Also it would make it easier for the filter removal to also remove the corresponding index which it should. So please coordinate with @kcover As removing the filter config should remove this object as well.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/spring/FilterIndexRepositoryIntegrationTest.java
```diff
@@ -0,0 +1,281 @@
...
+  public void testPojoPersistenceWhenIdIsNotDefined() {
+    exception.expect(UncategorizedSolrException.class);
+
+    final FilterIndexPojo pojo =
+        new FilterIndexPojo().setModifiedSince(MODIFIED_SINCE1).setFilterId(FILTER_ID1);
+
+    repo.save(pojo);
+  }
+
+  @Test
```
paouelle: ❗️ Add a test to verify that we can overwrite an existing index with a new one that changed its modified time

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/spring/FilterIndexRepositoryIntegrationTest.java
```diff
@@ -0,0 +1,281 @@
...
+  @Autowired private FilterIndexRepository repo;
+
+  @After
+  public void cleanup() {
+    repo.deleteAll();
+  }
+
+  @Test
+  public void testPojoPersistenceWhenIdIsNotDefined() {
+    exception.expect(UncategorizedSolrException.class);
```
paouelle: ❓ How do you know it failed because of the condition you are trying to create?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/spring/FilterIndexRepository.java
paouelle: Why? You are doing good :-D

replication-api/src/main/java/com/connexta/replication/api/data/FilterIndex.java
peterhuffer: Would there be a case with new fields in the future where we would instantiate an index before the first query?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
peterhuffer: I thought this pojo must support loading all versions? Could there ever be a minimum version less than the first version?

solr-schema/src/main/resources/solr/conf/replication-solr-managed-schema.xml
peterhuffer: We cannot right now because this schema applies to all collections so then every type would require this field.

replication-api/src/main/java/com/connexta/replication/api/data/FilterIndex.java
paouelle: 🤔 Maybe. So let's not make it require but do make it Optional<Instance> instead of returning `null`. The pojo would still return `null`

solr-schema/src/main/resources/solr/conf/replication-solr-managed-schema.xml
paouelle: good to know.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
peterhuffer: I still do not like having two persistent objects sharing the same UUID. I'd rather have a specific `removeByFilterId` on the FilterIndexRepository/Manager class to simplify removal of the FilterIndex.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/ReplicationItemManagerImpl.java
peterhuffer: Yes, but in this case we only care about the most recent since entities are sorted DESC by done time.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
paouelle: Unless you can guarantee that no 2 objects with different id but the same filter id can exist, I see no other ways. That check cannot be done in the code.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/ReplicationItemManagerImpl.java
paouelle: My point is that there should never be more than one.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/spring/FilterIndexRepositoryIntegrationTest.java
peterhuffer: Is that something we really need to verify? Can we not trust that solr has already tested the case of re-saving a config that already exists?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/ReplicationItemManagerImpl.java
peterhuffer: Yes there should be because an metadata can have multiple items. For instance if there is a create, update, delete on the same metadata. All 3 entries will have the same metadata_id field.

replication-api/src/main/java/com/connexta/replication/api/data/FilterIndex.java
peterhuffer: `Keeps track of where a filter is at when querying a Site.`?

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/spring/FilterIndexRepositoryIntegrationTest.java
paouelle: This is just to make sure that we setup the schema properly and all.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/ReplicationItemManagerImpl.java
paouelle: Sorry, I though I was looking at the filter index :-)

replication-api/src/main/java/com/connexta/replication/api/data/FilterIndex.java
paouelle: sure

replication-api-impl/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,6 @@
...
+<configuration>
+  <include resource="org/springframework/boot/logging/logback/base.xml"/>
+  <logger name="org.springframework" level="INFO"/>
+  <logger name="org.apache.solr" level="INFO" />
```
kcover: 👏 🥇 🎉

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexManagerImplTest.java
peterhuffer: Aren't I really testing the FilterIndexImpl readFrom ctor at that point in this test?

paouelle: If you want to mock the readFrom() to achieve it then fine. At the end of the day, you would be testing the contract of this constructor. The fact that it is currently implemented by calling readFrom() is beside the fact as we need to make sure it does what it is supposed to.

peterhuffer: Should we be testing the ctor of that class from within this manager test class

paouelle: Here you are not testing the ctor, you are just making sure that if the ctor/deserialization throws a persistence exception, it will come out from the find

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojo.java
peterhuffer: Left a todo for now until Kiefer's PR is merged.

peterhuffer:
build now

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/pojo/FilterIndexPojoTest.java
paouelle: ❗️ Still missing a test case for equals() when an object that is not a filter index is passed in

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterIndexImpl.java
kcover: This still needs to be checked for null right? If so don't forget to add a test for it.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterIndexImpl.java
```diff
@@ -0,0 +1,88 @@
...
+    readFrom(pojo);
+  }
+
+  @Override
+  public Optional<Instant> getModifiedSince() {
+    return Optional.ofNullable(modifiedSince);
+  }
+
+  @Override
+  public void setModifiedSince(Instant modifiedSince) {
```
paouelle: ❓ Since we are saying that the modified since is optional, shouldn't we have a setter that accepts `Optional<Instant>` as well?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterIndexImpl.java
paouelle: Actually no. Strike my comment above. It is optional now.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterIndexImpl.java
```diff
@@ -0,0 +1,88 @@
...
+              + " version: "
+              + pojo.getVersion()
+              + " for object: "
+              + getId());
+    } // do support pojo.getVersion() > CURRENT_VERSION for forward compatibility
+    readFromCurrentOrFutureVersion(pojo);
+  }
+
+  private void readFromCurrentOrFutureVersion(FilterIndexPojo pojo) {
+    setOrFailIfNull("modifiedSince", pojo::getModifiedSince, this::setModifiedSince);
```
paouelle: ❗️ I don't think we should be doing this check anymore since we are saying the modified since is optional.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterIndexImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+    when(filter.getId()).thenReturn(ID);
+  }
+
+  @Test
+  public void testDefaultCtor() {
+    FilterIndexImpl index = new FilterIndexImpl(filter);
+    assertThat(index.getModifiedSince(), OptionalMatchers.isEmpty());
+    assertThat(index.getId(), is(ID));
+  }
+
```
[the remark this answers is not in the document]
peterhuffer: It's in `testReadFrom`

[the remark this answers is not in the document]
paouelle: You should rely on the fact that the current ctor calls that method. You should have one method that tests the ctor and one that tests the readFrom().

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterIndexImpl.java
peterhuffer: No, if they pass in an Instant or null, the getter will handle the optional part but I should put nullable on the argument here.

paouelle: We should avoid using @nullable on API methods. It is preferable to overload it with an Optional flavor and leave the one there to not accept null.

peterhuffer: If I do that then I should check for null being passed in here and throw a runtime exception if so. Why not just let the caller pass in null, say for instance to reset the filter's modified since?

paouelle: no need to check for null since the prototype doesn't annotate the parameter with `@Nullable` which by convention means you should be passing it in.

It is a matter of keeping the API clean and avoid using `null` as a valid value which tends to create problems and complexity in client's code. This is one of the main reason why the `Optional` class was introduced

peterhuffer: Do we not want a way for the client to reset the modified time from the API?

paouelle: `setModifiedSince(Optional.empty())`

kcover:
Ran a full build locally including Itests and it passed! 🌮

PR #203, opened 2019-08-15
REPL-127 Re-introduce version support in all persisted data

paouelle:
#### What does this PR do?
This PR separates the JPA serialization requirements into private POJOs to allow better control over conversion when dealing with multiple versions.

It also move the affected classes to the new package scheme and renamed some for consistency.

Finally, it removes some of the setters from the data API which are no longer used to simply the interface for now.

@peterhuffer 
@clockard 
@kcover 

A build should be sufficient since it is only modifying the persistence layer and enough unit and integration test cases exist to cover all paths. In addition, the existing E2E test cases, will verify that the data is properly used by the replicator at runtime.

Fixes #XXXX

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests


replication-api/src/main/java/com/connexta/replication/api/impl/persistence/SiteManager.java
```diff
@@ -11,19 +11,19 @@
...
    *
-   * @param name the name to give the {@link ReplicationSite}
-   * @param url the for the {@link ReplicationSite}
-   * @return a new {@link ReplicationSite} implementation
+   * @param name the name to give the {@link Site}
+   * @param url the for the {@link Site}
+   * @return a new {@link Site} implementation
    */
-  ReplicationSite createSite(String name, String url);
+  Site create(String name, String url);
```
peterhuffer: ✏️ I don't think we are using this if we want to delete it.

replication-api/src/main/java/com/connexta/replication/api/impl/persistence/ReplicationItemManager.java
```diff
@@ -43,21 +43,14 @@
...
-  void saveItem(ReplicationItem replicationItem);
+  List<ReplicationItem> getAllForConfig(String configId, int startIndex, int pageSize);
 
   /**
    * Deletes all {@code ReplicationItem}s.
    *
    * @throws ReplicationPersistenceException RuntimeException thrown if delete was unsuccessful
    */
-  void deleteAllItems();
+  void removeAll();
```
peterhuffer: ❓ Do we use this anywhere? Can we delete it? This was a left over from the config remove command for easy cleanup.

replication-api/src/main/java/com/connexta/replication/api/data/Site.java
```diff
@@ -44,16 +37,9 @@
    *
    * @param url the URL to give this site
    */
+  // still internally used by the ReplicatorImpl
```
peterhuffer: If we want to get rid of these now, we can just copy the object with the updated values and re-save it with the same ID and the original record will get overwritten.

replication-api/src/main/java/com/connexta/replication/api/data/Persistable.java
```diff
@@ -22,18 +22,4 @@
...
-   * @param id the ID to set as the persistable object's ID
-   */
-  void setId(String id);
-
-  /**
-   * Returns the version of this persistable
-   *
-   * @return integer version of the persistable
-   */
-  int getVersion();
```
peterhuffer: ❓ This was remove because the object always represents the latest version, right?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/SitePojo.java
```diff
@@ -0,0 +1,170 @@
...
+@SolrDocument(collection = "replication_site")
+public class SitePojo extends Pojo<SitePojo> {
+  /**
+   * List of possible versions:
+   *
+   * <ul>
+   *   <li>1 - initial version.
+   * </ul>
+   */
+  public static final int CURRENT_VERSION = 2;
```
peterhuffer: ❓ This should be version 1 right?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -0,0 +1,178 @@
...
+   * Serializes this object to the specified pojo that can be serialized to a spring-data supported
+   * data store. Any implementation of this method in a subclass should first make a call to the
+   * super version of this method before serializing its own properties.
+   *
+   * @param pojo the pojo to serialize this object to
+   * @return <code>pojo</code> for chaining
+   * @throws ReplicationPersistenceException if an error occurs while trying to serialize this
+   *     object
+   */
+  protected P writeTo(P pojo) {
```
peterhuffer: ❓ Do we need to set the version here?

replication-api/src/main/java/com/connexta/replication/api/impl/persistence/SiteManager.java
paouelle: They are currently in the `DataManager` interface so I kept them for the time being.

replication-api/src/main/java/com/connexta/replication/api/impl/persistence/ReplicationItemManager.java
paouelle: I am all for removing it.

replication-api/src/main/java/com/connexta/replication/api/data/Persistable.java
paouelle: `getVersion()` is for the schema version and not the object version 😉

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/SitePojo.java
paouelle: yup. good catch

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
paouelle: I am on the fence. I am assuming that we are always serializing to the latest version which is expected to be set by the constructor of the pojo. On one side, I am not a fan on relying on something else to do it. On the other side, I didn't want the persistable to be aware of the version. But since they are aware since they are responsible for converting, I will introduce it in each specific persistable `writeToPojo)_` classes.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/ReplicationItemManagerTest.java
```diff
@@ -85,19 +84,20 @@ public void getItemsForConfig() {
...
-
-  @Test
-  public void saveItem() {
-    ReplicationItemImpl item = new ReplicationItemImpl();
-    itemManager.saveItem(item);
-    verify(itemRepository).save(eq(item));
+    assertThat(itemManager.getAllForConfig("configId", 0, 1), is(items));
   }
 
+  // BOB: redefine this one
```
clockard: Should this just be removed or are you coming back to it?

kcover:
❓ Is the idea to have the pojos contain all the fields of every version and just pick out the ones we need?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -0,0 +1,178 @@
...
+   * @param field the name of the field being check for presence (used in the exception message)
+   * @param supplier a supplier capable of retrieving the pojo's current value for the field
+   * @param consumer a consumer capable of updating this object with the field's value if it is
+   *     defined
+   * @param <F> the field's type
+   * @throws ReplicationPersistenceException if the field's value as supplied by <code>supplier
+   *     </code> is <code>null</code>
+   */
+  protected <F> void setOrFailIfNull(
+      String field, Supplier<? extends F> supplier, Consumer<? super F> consumer) {
```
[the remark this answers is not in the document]
clockard: I was wondering the same thing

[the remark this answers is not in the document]
paouelle: Really good question. I guess I started with the need for a consumer and decided to keep it consistent. It also allows me to potentially control potential exceptions in the future while retrieving (although currently I don't need to do that).  Since this method is intended to validate the value before setting it, I saw no reasons for allowing you to pass a constant since you wouldn't need to validate it and as such, could have set it directly without the need for this method.

I am good either way.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/ReplicationItemPojo.java
```diff
@@ -0,0 +1,404 @@
...
+  /**
+   * List of possible versions:
+   *
+   * <ul>
+   *   <li>1 - initial version.
+   * </ul>
+   */
+  public static final int CURRENT_VERSION = 1;
+
+  @Indexed(name = "metadata_id", type = "string")
```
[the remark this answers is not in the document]
paouelle: Wasn't the type already defined in the original class? I simply copied that over. As for being unsearchable, I guess simply because I didn't think we would ever need to search based on them. I can definitely revert that if you think we would need to search for those particular fields.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -0,0 +1,178 @@
...
+  protected final String type;
+
+  private String id;
+
+  protected AbstractPersistable(String type) {
+    this.type = type;
+    this.id = UUID.randomUUID().toString();
+  }
+
+  protected AbstractPersistable(String type, @Nullable String id) {
```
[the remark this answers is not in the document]
paouelle: Why? Shouldn't all objects we persist have a unique internal identifier? 
Our original code had that restriction so I am not changing that.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/ReplicationItemManagerTest.java
paouelle: I am coming back to it as part of fixing and adding unit tests. `BOB` is my personal `TODO` :-)

paouelle:
@kcover Yup. Similar to how we did it with maps on the release branch. There the map would contain whatever was serialized for the given version. Here, we are forced to have a place for spring-data to deserialize the fields it sees in Solr. So we are forced to keep around all fields for all versions we are supporting at one given time.

replication-api/src/main/java/com/connexta/replication/api/data/Site.java
paouelle: Didn't want to do that as part of this PR especially since some of that might/will change as part of our redesign. I just cleaned up what I could for now and left the rest such that it wouldn't break things for now.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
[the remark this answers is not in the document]
peterhuffer: I think we should enforce and ID. We had this issue with Replication Items relying on metadata ID as the primary ID and duplicate data unknowingly overwritting each other.

replication-api/src/main/java/com/connexta/replication/api/data/Persistable.java
```diff
@@ -22,18 +22,4 @@
    * @return The ID of the object
    */
   String getId();
```
peterhuffer: ❓ Sometimes is it useful to track the last time a configuration was modified and the user who modified it. Does anyone see any value in adding these 2 fields? I don't know if we would add them here or on an as needed basis per configuration.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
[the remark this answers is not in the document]
paouelle: @peterhuffer Was this a reply to Chris or a comment to the existing code? I believe the existing code does enforce the presence of a unique internal id. The first constructor is meant to be used when a new object is created and the other when we are deserializing an existing one. I just added javadoc for that and updated the subclass constructors to call this one with a `null` id since the id is restored b y the `readFrom()`

replication-api/src/main/java/com/connexta/replication/api/data/Persistable.java
paouelle: I agree that these are a very good thing to have for traceability. 

I think we can wait before adding then since we currently do not have a UI or a way to create/update these objects from a user. As such, it might provide little value (aside maybe for the last modified timestamp).

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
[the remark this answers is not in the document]
peterhuffer: In reply to Chris' comment.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/ReplicationItemPojo.java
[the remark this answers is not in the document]
kcover: Does the query by configId and status in the itemRepo class still work if status isn't searchable? Other than that I can't think of a specific reason we would search for another field so I think it's fine to leave them as is.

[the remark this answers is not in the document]
paouelle: Nice catch. This is exactly why I am working on integration test cases :-)

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/ReplicatorConfigImpl.java
```diff
@@ -141,4 +97,65 @@ public void setLastMetadataModified(Date lastMetadataModified) {
...
+  }
+
+  @Override
+  protected void readFrom(ReplicatorConfigPojo pojo) {
+    super.readFrom(pojo);
+    setOrFailIfNullOrEmpty("name", pojo::getName, this::setName);
+    setOrFailIfNullOrEmpty("source", pojo::getSource, this::setSource);
+    setOrFailIfNullOrEmpty("destination", pojo::getName, this::setDestination);
+    setOrFailIfNullOrEmpty("filter", pojo::getFilter, this::setFilter);
+    this.description = pojo.getDescription();
```
kcover: Should description be marked nullable, since we're letting it be null or empty?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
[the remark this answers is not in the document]
clockard: My comment was related to the @Nullable which I don't think we should allow. We should always have an id. If we let this be nullable we should at least call the this(type) so we generate a new one.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/EmbeddedSolrServerFactory.java
```diff
@@ -0,0 +1,102 @@
...
+import org.apache.commons.io.FileUtils;
+import org.apache.solr.client.solrj.embedded.EmbeddedSolrServer;
+import org.apache.solr.core.CoreContainer;
+import org.springframework.beans.factory.DisposableBean;
+import org.springframework.data.solr.server.SolrClientFactory;
+import org.springframework.util.Assert;
+import org.springframework.util.ResourceUtils;
+
+/**
+ * This class provides an implementation for creating an embedded Solr server factory. If will
```
clockard: `It will`

replication-api/src/main/java/com/connexta/replication/api/impl/persistence/DataManager.java
```diff
@@ -11,9 +11,9 @@
  * License is distributed along with this program and can be found at
  * <http://www.gnu.org/licenses/lgpl.html>.
  */
-package com.connexta.ion.replication.api.persistence;
+package com.connexta.replication.api.impl.persistence;
```
clockard: Why did we add the impl to the package name? This is still an interface in the api module.

replication-api/src/main/java/com/connexta/replication/api/impl/persistence/ReplicationItemManager.java
```diff
@@ -11,17 +11,17 @@
  * License is distributed along with this program and can be found at
  * <http://www.gnu.org/licenses/lgpl.html>.
  */
-package com.connexta.ion.replication.api.persistence;
+package com.connexta.replication.api.impl.persistence;
```
clockard: Same comment about the impl package

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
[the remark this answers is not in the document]
paouelle: It is a protected constructor used  by derived classes when deserializing to avoid initializing an id before the `readFrom()` is called which does that.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/ReplicationItemManagerTest.java
paouelle: Removed

distributions/application/src/main/java/com/connexta/ion/replication/application/Main.java
```diff
@@ -25,8 +25,8 @@
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.data.solr.repository.config.EnableSolrRepositories;
 
-@SpringBootApplication(scanBasePackages = {"com.connexta.ion.replication"})
-@EnableSolrRepositories(basePackages = "com.connexta.ion.replication")
+@SpringBootApplication(scanBasePackages = "com.connexta.ion.replication")
```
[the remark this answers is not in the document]
paouelle: just because not everything has moved over yet

distributions/docker/solr/src/main/docker/files.xml
```diff
@@ -15,15 +15,24 @@
...
-      </includes>
-      <outputDirectory>.</outputDirectory>
-    </fileSet>
-  </fileSets>
+    <!-- This assembly descriptor is used to provide files and dependencies
+      to the docker build context. -->
+    <dependencySets>
+        <dependencySet>
+            <includes>
+                <include>*</include>
```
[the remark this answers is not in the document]
paouelle: It does copy more than needed. However, the docker file only copies to the docker instance the schema. This was much easier than trying to hardcode yet in another place the name of the file we needed.

pom.xml
```diff
@@ -74,8 +75,8 @@
         <maven.compiler.source>11</maven.compiler.source>
         <maven.compiler.target>11</maven.compiler.target>
         <java.version>1.11</java.version>
-        <snapshots.repository.url/>
-        <releases.repository.url/>
+        <snapshots.repository.url />
```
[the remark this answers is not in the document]
paouelle: the formatter does that autoamtically

pom.xml
```diff
@@ -128,6 +131,7 @@
         <micrometer.version>1.2.0</micrometer.version>
         <javax-validation-api.version>2.0.1.Final</javax-validation-api.version>
         <httpclient.version>4.5.9</httpclient.version>
+        <to-string-verifier.version>1.4.5</to-string-verifier.version>
```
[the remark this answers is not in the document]
paouelle: No idea about that one. This came in when I rebased

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/EmbeddedSolrServerFactory.java
paouelle: `it` you say so 😄

replication-api/src/main/java/com/connexta/replication/api/impl/persistence/DataManager.java
paouelle: oops!

paouelle: I think IntelliJ played a joke on me

paouelle:
Not sure I want the Integration tests to be skippable with -DskipITs. This was used for more complex test cases. Further more, I want to wait for the guidelines to be defined before renaming things. For now, I just named them with `Integration` to differentiate them.

As for the logs, I would also like to reduce that. Couldn't figure out how to do it so I left it as is. If you can turn them down then go ahead (please). I don't think we should hold off this PR for that right now.

PR #208, opened 2019-08-19
Fix erroneous file copies of global dependencies

peterhuffer:
#### What does this PR do?
- Fix erroneous file copies of global dependencies that were being copied into target

@clockard @paouelle @kcover 

CI

### DIFF
--- a/distributions/docker/replication/src/main/docker/files.xml
+++ b/distributions/docker/replication/src/main/docker/files.xml
@@ -20,7 +20,7 @@
   <dependencySets>
     <dependencySet>
       <includes>
-        <include>*</include>
+        <include>com.connexta.ion.replication.distributions:replication-distros-spring</include>
       </includes>
       <outputDirectory>.</outputDirectory>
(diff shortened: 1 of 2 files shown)


distributions/docker/replication/src/main/docker/files.xml
```diff
@@ -20,7 +20,7 @@
   <dependencySets>
     <dependencySet>
       <includes>
-        <include>*</include>
+        <include>com.connexta.ion.replication.distributions:replication-distros-spring</include>
       </includes>
       <outputDirectory>.</outputDirectory>
       <useTransitiveDependencies>false</useTransitiveDependencies>
```
peterhuffer: Below this where we copy resources, is there a reason that we copy the demo keystore/truststore over to the target directory? I was just thinking of deleting this assembly file.

clockard: We copy them over so the docker image will work "out of the box" if you don't configure anything externally.

PR #211, opened 2019-08-20
REP-207 Add filter config

### LINKED ISSUE #207: Create a new standalone Filter configuration which a Site can use to describe what data it wants
**Describe the task**
A new Solr core needs to be created that holds filter information for sites. This will be similar to the current replication-configuration core but will have only the following fields.

    id
    version
    related site id
    filter (cql string)
    name
    description
    suspended (boolean)

A manager to read/write this core also needs to be created.

kcover:
#### What does this PR do?
Adds a filter config object to replication along with the persistence manager and tests to go with it.

@peterhuffer @clockard @paouelle 

CI Build 

Fixes #207 

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterImpl.java
+++ b/replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterImpl.java
@@ -1,0 +1,156 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.api.impl.data;
+
+import com.connexta.ion.replication.api.ReplicationPersistenceException;
+import com.connexta.replication.api.data.Filter;
+import com.connexta.replication.api.impl.persistence.pojo.FilterPojo;
+import com.google.common.annotations.VisibleForTesting;
+import java.util.Optional;
+
+/**
+ * A filter describing what data to replicate to a site. A site can have multiple filters associated
+ * with it to describe what data it would like to receive.
+ */
+public class FilterImpl extends AbstractPersistable<FilterPojo> implements Filter {
+  private static final String TYPE = "replication filter";
+
+  private String siteId;
+
+  private String filter;
+
+  private String name;
+
+  private String description;
+
+  private boolean isSuspended;
+
+  private byte priority;
+
+  /** Creates a default filter. */
+  public FilterImpl() {
+    super(FilterImpl.TYPE);
+  }
+
+  /**
+   * Creates a filter from the given {@link FilterPojo}.
+   *
+   * @param pojo a pojo containing the values this filter should be instantiated with.
+   */
+  protected FilterImpl(FilterPojo pojo) {
+    super(FilterImpl.TYPE);
+    readFrom(pojo);
+  }
+
+  @Override
+  public String getSiteId() {
+    return siteId;
+  }
+
+  @Override
+  public String getFilter() {
+    return filter;
+  }
+
+  @Override
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public Optional<String> getDescription() {
+    return Optional.ofNullable(description);
+  }
+
+  @Override
+  public boolean isSuspended() {
+    return isSuspended;
+  }
+
+  @Override
+  public byte getPriority() {
+    return priority;
+  }
+
+  @VisibleForTesting
+  void setSiteId(String siteId) {
+    this.siteId = siteId;
+  }
+
+  @VisibleForTesting
+  void setFilter(String filter) {
+    this.filter = filter;
+  }
+
+  @VisibleForTesting
+  void setName(String name) {
+    this.name = name;
+  }
+
+  @VisibleForTesting
+  void setDescription(String description) {
+    this.description = description;
+  }
+
+  @VisibleForTesting
+  void setSuspended(boolean suspended) {
+    this.isSuspended = suspended;
+  }
+
+  @VisibleForTesting
+  void setPriority(byte priority) {
+    if (priority < 1 || priority > 10) {
+      throw new IllegalArgumentException(
+          "Filter priority must be between 1 and 10 but was " + priority);
+    } else {
+      this.priority = priority;
+    }
+  }
+
+  @Override
+  protected FilterPojo writeTo(FilterPojo pojo) {
+    super.writeTo(pojo);
+    setOrFailIfNullOrEmpty("siteId", this::getSiteId, pojo::setSiteId);
+    setOrFailIfNullOrEmpty("filter", this::getFilter, pojo::setFilter);
+    setOrFailIfNullOrEmpty("name", this::getName, pojo::setName);
+    return pojo.setVersion(FilterPojo.CURRENT_VERSION)
+        .setDescription(description)
+        .setSuspended(isSuspended)
+        .setPriority(priority);
+  }
+
+  @Override
+  protected void readFrom(FilterPojo pojo) {
+    super.readFrom(pojo);
+    if (pojo.getVersion() < FilterPojo.MINIMUM_VERSION) {
+      throw new ReplicationPersistenceException(
+          "unsupported "
+              + FilterImpl.TYPE
+              + " version: "
+              + pojo.getVersion()
+              + " for object: "
+              + getId());
+    } // do support pojo.getVersion() > CURRENT_VERSION for forward compatibility
+    readFromCurrentOrFutureVersion(pojo);
+  }
+
+  private void readFromCurrentOrFutureVersion(FilterPojo pojo) {
+    setOrFailIfNullOrEmpty("siteId", pojo::getSiteId, this::setSiteId);
+    setOrFailIfNullOrEmpty("filter", pojo::getFilter, this::setFilter);
+    setOrFailIfNullOrEmpty("name", pojo::getName, this::setName);
+    this.description = pojo.getDescription();
+    this.isSuspended = pojo.isSuspended();
+    this.priority = (byte) Math.min(Math.max(pojo.getPriority(), 1), 10);
+  }
+}
--- a/replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterManagerImpl.java
+++ b/replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterManagerImpl.java
@@ -1,0 +1,77 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.api.impl.data;
+
+import com.connexta.ion.replication.api.NotFoundException;
+import com.connexta.replication.api.data.Filter;
+import com.connexta.replication.api.impl.persistence.pojo.FilterPojo;
+import com.connexta.replication.api.impl.persistence.spring.FilterRepository;
+import com.connexta.replication.api.persistence.FilterManager;
+import java.util.Spliterator;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/** Performs CRUD operations for {@link Filter}s. */
+public class FilterManagerImpl implements FilterManager {
+
+  private final FilterRepository filterRepository;
+
+  public FilterManagerImpl(FilterRepository filterRepository) {
+    this.filterRepository = filterRepository;
+  }
+
+  @Override
+  public Filter create() {
+    return new FilterImpl();
+  }
+
+  @Override
+  public Filter get(String id) {
+    return filterRepository
+        .findById(id)
+        .map(FilterImpl::new)
+        .orElseThrow(
+            () -> new NotFoundException(String.format("Cannot find filter with ID: %s", id)));
+  }
+
+  @Override
+  public Stream<Filter> objects() {
+    return filterStreamOf(filterRepository.findAll().spliterator());
+  }
+
+  @Override
+  public void save(Filter filter) {
+    if (filter instanceof FilterImpl) {
+      filterRepository.save(((FilterImpl) filter).writeTo(new FilterPojo()));
+    } else {
+      throw new IllegalArgumentException(
+          "Expected a FilterImpl but got a " + filter.getClass().getSimpleName());
+    }
+  }
+
+  @Override
+  public void remove(String id) {
+    filterRepository.deleteById(id);
+  }
+
+  @Override
+  public Stream<Filter> filtersForSite(String siteId) {
+    return filterStreamOf(filterRepository.findBySiteId(siteId).spliterator());
+  }
+
+  // coverts the pojos to Filters and returns them as a stream
+  private Stream<Filter> filterStreamOf(Spliterator<FilterPojo> filterPojos) {
+    return StreamSupport.stream(filterPojos, false).map(FilterImpl::new).map(Filter.class::cast);
+  }
+}
--- a/replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojo.java
+++ b/replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojo.java
@@ -1,0 +1,216 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.api.impl.persistence.pojo;
+
+import java.util.Objects;
+import javax.annotation.Nullable;
+import org.springframework.data.solr.core.mapping.Indexed;
+import org.springframework.data.solr.core.mapping.SolrDocument;
+
+/** A pojo used to load filters from persistence. */
+@SolrDocument(collection = FilterPojo.COLLECTION)
+public class FilterPojo extends Pojo<FilterPojo> {
+
+  /**
+   * List of possible versions:
+   *
+   * <ul>
+   *   <li>1 - initial version.
+   * </ul>
+   */
+  public static final int CURRENT_VERSION = 1;
+
+  /** The oldest version supported by the current code (anything before that will fail). */
+  public static final int MINIMUM_VERSION = 1;
+
+  public static final String COLLECTION = "replication_filter";
+
+  @Indexed(name = "site_id")
+  @Nullable
+  private String siteId;
+
+  @Indexed(name = "filter", searchable = false)
+  @Nullable
+  private String filter;
+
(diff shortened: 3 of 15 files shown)


replication-api/src/main/java/com/connexta/replication/api/data/Filter.java
```diff
@@ -0,0 +1,56 @@
...
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.api.data;
+
+import javax.annotation.Nullable;
+
+/** A filter for retrieving data from a specific site. */
+public interface Filter extends Persistable {
```
clockard: Should filters also have a priority? When we are putting items in our priority queues I would imagine that the filter priority would be used as the item priority in the queue.

kcover: Is that going to be used atm? I think we'll want to do it eventually but I'm not sure if it should be done right now. I could just add a priority that should be a number 1-10 if we know that's how we want to do it.

clockard: I think it will be used when we switch over to queues in a couple sprints. We could wait till then but that's just one more thing to do and potentially a different version of the config.

clockard: The 1-10 matches up with the queue broker api priority ranges so I think that would be ok. The other option would be to use the enums defined in the api but I think i like just a straight 1-10 better.

paouelle: I would prefer we be consistent and reuse the `PriorityLevel` enum by moving it into a common package to be shared by both the API and the queue.

If not then we should remove that enum and opt for a 1-10 everywhere.

Either way I opt for consistency.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojo.java
```diff
@@ -0,0 +1,189 @@
...
+          && Objects.equals(name, pojo.name)
+          && Objects.equals(description, pojo.description);
+    }
+    return false;
+  }
+
+  @Override
+  public String toString() {
+    return String.format(
+        "FilterPojo[id=%s, version=%s, siteId=%s, filter=%s, name=%s, description=%s]",
```
paouelle: ✏️ missing `suspended`

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterImpl.java
```diff
@@ -0,0 +1,119 @@
...
+    return filter;
+  }
+
+  @Override
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public String getDescription() {
```
paouelle: ✏️ Should return `Optional<String>`

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterImpl.java
```diff
@@ -0,0 +1,119 @@
...
+    return description;
```
paouelle:
```suggestion
    return Optional.ofNullable(description);
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterImpl.java
```diff
@@ -0,0 +1,119 @@
...
+    return pojo.setSiteId(siteId)
+        .setFilter(filter)
+        .setName(name)
+        .setDescription(description)
+        .setSuspended(isSuspended);
+  }
+
+  @Override
+  protected void readFrom(FilterPojo pojo) {
+    super.readFrom(pojo);
```
paouelle: ❗️ Add the following after `super.readFrom()`
```
if (pojo.getVersion() < FilterPojo.MINIMUM_VERSION) {
      throw new ReplicationPersistenceException(
          "unsupported "
              + FilterImpl.TYPE
              + " version: "
              + pojo.getVersion()
              + " for object: "
              + getId());
    } // do support pojo.getVersion() > CURRENT_VERSION for forward compatibility
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterImpl.java
```diff
@@ -0,0 +1,119 @@
...
+    setOrFailIfNullOrEmpty("siteId", pojo::getSiteId, this::setSiteId);
```
paouelle: ✏️ I ended up moving the remainder of this method into a method named `readFromCurrentAndFutureVersion(FilterPojo pojo)` and called it from it in preparation of future versions support

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterManagerImpl.java
```diff
@@ -0,0 +1,77 @@
...
+    return filterStreamOf(filterRepository.findAll().spliterator());
+  }
+
+  @Override
+  public void save(Filter filter) {
+    if (filter instanceof FilterImpl) {
+      filterRepository.save(((FilterImpl) filter).writeTo(new FilterPojo()));
+    } else {
+      throw new IllegalArgumentException(
+          "Expected a ReplicatorConfigImpl but got a " + filter.getClass().getSimpleName());
```
paouelle: `FilterImpl`

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojoTest.java
```diff
@@ -0,0 +1,139 @@
...
+  @Test
+  public void hashcode() {
+    FilterPojo filterPojo2 = makeDefaultPojo();
+    assertThat(filterPojo.hashCode(), is(filterPojo2.hashCode()));
+    filterPojo2.setName("numbatwo");
+    assertThat(filterPojo.hashCode(), not(filterPojo2.hashCode()));
+  }
+
+  @Test
+  public void testToString() {
```
paouelle: ✏️ Not sure we should test the `toString()`

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojoTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+  private static final String NAME = "name";
+
+  private static final String DESCRIPTION = "description";
+
+  private FilterPojo filterPojo;
+
+  @Before
+  public void setup() {
+    filterPojo = makeDefaultPojo();
```
paouelle: ❓ Why do this in a setup method? Why not just initialize the attribute above? By design JUnit will create a new instance for each test it runs

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterManagerImplTest.java
```diff
@@ -0,0 +1,165 @@
...
+  @Test
+  public void save() {
+    FilterPojo filterPojo = makeDefaultPojo();
+    filterManager.save(new FilterImpl(filterPojo));
+    verify(filterRepository).save(eq(filterPojo));
+  }
+
+  @Test(expected = IllegalArgumentException.class)
+  public void saveIllegalObject() {
+    filterManager.save(new TestFilter());
```
paouelle: ❓ Can we not just use Mockito to create a mock of `Filter`?

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+    filter.setSuspended(true);
+    assertThat(filter.getSiteId(), is(SITE_ID));
+    assertThat(filter.getFilter(), is(FILTER));
+    assertThat(filter.getName(), is(NAME));
+    assertThat(filter.getDescription(), is(DESCRIPTION));
+    assertThat(filter.isSuspended(), is(true));
+  }
+
+  @Test
+  public void readFrom() {
```
paouelle: ❗️ We should add test cases to verify the failures that can occur when we read and validate the data we receive

replication-api/src/main/java/com/connexta/replication/api/data/Filter.java
clockard: Agree we should try to be consistent. I think the 1-10 gives us more flexibility but I am ok either way. I do think we should include the priority in this iteration though.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterImplTest.java
```diff
@@ -0,0 +1,88 @@
...
+    filter = new FilterImpl(pojo);
+    assertThat(filter.getSiteId(), is(SITE_ID));
+    assertThat(filter.getFilter(), is(FILTER));
+    assertThat(filter.getName(), is(NAME));
+    assertThat(filter.getDescription(), is(DESCRIPTION));
+    assertThat(filter.isSuspended(), is(true));
+  }
+
+  @Test
+  public void writeTo() {
```
paouelle: ❗️ We should add test cases to verify the failures that can occur when we write and validate the data

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/spring/FilterRepositoryIntegrationTest.java
```diff
@@ -0,0 +1,368 @@
...
+    repo.save(POJO2);
+    repo.save(POJO3);
+
+    assertThat(repo.count(), equalTo(3L));
+    repo.deleteAll();
+    assertThat(repo.count(), equalTo(0L));
+  }
+
+  @Test
+  public void testFindBySiteId() {
```
paouelle: ❗️ Add another pojo with `SITE_ID` to verify you get more than one match back.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterManagerImplTest.java
kcover: I need the instanceof statement to be false. I wasn't sure how to achieve that with a mock.

solr-schema/src/main/resources/solr/conf/replication-solr-managed-schema.xml
```diff
@@ -165,17 +165,17 @@
...
-    <field name="suspended" type="boolean" indexed="true" stored="true" required="false"
-           multiValued="false" />
     <field name="last_metadata_modified" type="pdate" indexed="true" stored="false" required="false"
            multiValued="false" />
 
+    <!-- replication_filter fields -->
+    <field name="name" type="string" indexed="true" stored="true" required="false" multiValued="false" />
+    <field name="site_id" type="string" indexed="true" stored="true" required="false" multiValued="false" />
+    <field name="filter" type="string" indexed="true" stored="true" required="false" multiValued="false" />
+    <field name="suspended" type="boolean" indexed="true" stored="true" required="false" multiValued="false" />
```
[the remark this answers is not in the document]
kcover: Now that I think about it I don't think so, I'll change this and filter to not be indexed and make them unsearchable in the impls.

replication-api/src/main/java/com/connexta/replication/api/data/Filter.java
paouelle: Let's go with a byte and fix it to 1-10. Let's make the change to the queue api to remove the enum and set it as a byte there as well. This should be a simple change that we could do at the same time for consistency.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterManagerImplTest.java
paouelle: Mock the interface. This will dynamically create an object class that implements it but that will be different than the one internally.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojoTest.java
kcover: I thought I would be changing the values of filterPojo in tests when I wrote this. Turns out I didn't have to. I'll change it

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojo.java
peterhuffer: ✏️ Can add a toString test and use the ToStringVerifier to catch missing fields like this.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojo.java
```diff
@@ -0,0 +1,189 @@
...
+    return siteId;
+  }
+
+  /**
+   * Sets the site id for this filter
+   *
+   * @param siteId the id of the site to associate with this filter.
+   * @return this for chaining
+   */
+  public FilterPojo setSiteId(String siteId) {
```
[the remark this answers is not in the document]
kcover: Except for suspended and priority right?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterImpl.java
kcover: should that be `readFromCurrentAndFutureVersions` or `readFromCurrentOrFutureVersion`? I'm thinking the latter

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojo.java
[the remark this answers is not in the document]
paouelle: yeah. all the object-based ones, not the primitive :-)

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterImpl.java
paouelle: or ... `readFromCurrentOrFutureVersionAsLongAsItIsGreaterThanTheMinimumSupportedVersion()`

😄 

I like the later. @peterhuffer Take note of the method name change

solr-schema/src/main/resources/solr/conf/replication-solr-managed-schema.xml
```diff
@@ -173,8 +173,9 @@
     <!-- replication_filter fields -->
     <field name="name" type="string" indexed="true" stored="true" required="false" multiValued="false" />
     <field name="site_id" type="string" indexed="true" stored="true" required="false" multiValued="false" />
-    <field name="filter" type="string" indexed="true" stored="true" required="false" multiValued="false" />
-    <field name="suspended" type="boolean" indexed="true" stored="true" required="false" multiValued="false" />
+    <field name="filter" type="string" indexed="false" stored="true" required="false" multiValued="false" />
+    <field name="suspended" type="boolean" indexed="false" stored="true" required="false" multiValued="false" />
+    <field name="priority" type="pint" indexed="true" stored="true" required="false" multiValued="false" />
```
kcover: We should not be indexing this.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterImpl.java
```diff
@@ -0,0 +1,167 @@
...
+    readFromCurrentOrFutureVersion(pojo);
+  }
+
+  private void readFromCurrentOrFutureVersion(FilterPojo pojo) {
+    setOrFailIfNullOrEmpty("siteId", pojo::getSiteId, this::setSiteId);
+    setOrFailIfNullOrEmpty("filter", pojo::getFilter, this::setFilter);
+    setOrFailIfNullOrEmpty("name", pojo::getName, this::setName);
+    this.description = pojo.getDescription();
+    this.isSuspended = pojo.isSuspended();
+    this.priority = readPriority(pojo);
```
paouelle:
```suggestion
    this.priority = Math.min(Math.max(pojo.getPriority(), 1), 10);
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/FilterManagerImpl.java
```diff
@@ -0,0 +1,77 @@
...
+import com.connexta.replication.api.impl.persistence.spring.FilterRepository;
+import com.connexta.replication.api.persistence.FilterManager;
+import java.util.Spliterator;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/** Performs CRUD operations for {@link Filter}s. */
+public class FilterManagerImpl implements FilterManager {
+
+  FilterRepository filterRepository;
```
paouelle: ✏️ private and final

replication-api-impl/src/main/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojo.java
```diff
@@ -0,0 +1,216 @@
...
+          && Objects.equals(name, pojo.name)
+          && Objects.equals(description, pojo.description);
+    }
+    return false;
+  }
+
+  @Override
+  public String toString() {
+    return String.format(
+        "FilterPojo[id=%s, version=%s, siteId=%s, filter=%s, name=%s, description=%s, suspended=%s, priority=%s]",
```
paouelle: ✏️ `%d` for priority

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterManagerImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+    assertThat(filter.getName(), is(NAME));
+    assertThat(filter.getPriority(), is(PRIORITY));
+  }
+
+  @Test(expected = NotFoundException.class)
+  public void getFilterNotFound() {
+    when(filterRepository.findById(anyString())).thenReturn(Optional.empty());
+    filterManager.get("id");
+  }
+
```
paouelle: ❗️ We should add a test case when it fails because the persitence layer throws a persistence exception

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterManagerImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+  @Test
+  public void objects() {
```
paouelle: ❗️ We should add a test case when it fails because the persitence layer throws a persistence exception on one of the object

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterManagerImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+    assertThat(filter.getId(), is(ID));
+    assertThat(filter.getSiteId(), is(SITE_ID));
+    assertThat(filter.getDescription().get(), is(DESCRIPTION));
+    assertThat(filter.getFilter(), is(FILTER));
+    assertThat(filter.getName(), is(NAME));
+    assertThat(filter.getPriority(), is(PRIORITY));
+  }
+
+  @Test
+  public void save() {
```
paouelle: ❗️ We should add a test case when it fails because the persitence layer throws a persistence exception

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterManagerImplTest.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  @Test
+  public void remove() {
+    filterManager.remove("id");
+    verify(filterRepository).deleteById("id");
+  }
+
+  @Test
+  public void filtersForSite() {
```
paouelle: ❗️ We should add a test case when it fails because the persitence layer throws a persistence exception on one of the object

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/pojo/FilterPojoTest.java
```diff
@@ -0,0 +1,142 @@
...
+
+  @Test
+  public void equalsWithDifferentPriority() {
+    FilterPojo filterPojo2 = makeDefaultPojo().setPriority((byte) 5);
+    assertFalse(filterPojo.equals(filterPojo2));
+    assertFalse(filterPojo2.equals(filterPojo));
+  }
+
+  @Test
+  public void equalsWithDifferentObject() {
```
paouelle: we should add a test when called with null

replication-api/src/main/java/com/connexta/replication/queue/TaskInfo.java
```diff
@@ -20,11 +20,11 @@
 /** Represents a task that can be queued for later processing by a worker. */
 public interface TaskInfo {
   /**
-   * Gets the priority for this task.
+   * Gets the priority for this task, represented as a number between 1 and 10.
    *
    * @return the priority for the task
    */
-  public PriorityLevel getPriority();
```
paouelle: ❗️ Remove the PriorityLevel enum

replication-api-impl/src/test/java/com/connexta/replication/api/impl/data/FilterManagerImplTest.java
kcover: You mean the repository right? Why would we test the repository here?

replication-api/src/main/java/com/connexta/replication/queue/TaskInfo.java
kcover: I did apparently git thinks I changed it to be the filterManager interface.

PR #215, opened 2019-08-22
REP-214 Add metadata to Ion create requests

### LINKED ISSUE #214: Add metadata to Ion create requests
**Describe the task**
Include the metacard xml as a separate attachment when performing create requests with the ion adapter.

clockard:
#### What does this PR do?
Adds metadata to Ion create requests
**Attachment name still needs approval from CDR team**
@peterhuffer @paouelle @kcover 

Manual test in integration environment

Fixes #214

### DIFF
--- a/adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/DdfMetadata.java
+++ b/adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/DdfMetadata.java
@@ -1,0 +1,53 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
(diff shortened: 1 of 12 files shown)


adapters/ion-adapter/src/main/java/com/connexta/ion/replication/adapters/ion/IonNodeAdapter.java
```diff
@@ -150,6 +152,18 @@ public void close() throws IOException {
     // nothing to close
   }
 
+  private HttpEntity getMetadataEntity(Metadata metadata) {
```
clockard: The get entity methods were added to ensure the content types are set correctly.

adapters/ion-adapter/src/main/java/com/connexta/ion/replication/adapters/ion/IonNodeAdapterFactory.java
```diff
@@ -58,7 +58,6 @@ public IonNodeAdapterFactory(
 
   @Override
   public NodeAdapter create(URL url) {
-    // TODO change this when ingest enpoint supports https
```
paouelle: ❓ Why is this no longer a todo?

clockard: It was getting flagged by sonar cloud and there is nothing we could do about it. When the ingest endpoint updates to use https they will notify us and we can make the change then. Don't think we need to the todo to track that.

PR #229, opened 2019-08-29
REP-225 Add support for query service and manager

### LINKED ISSUE #225: Add support for a query service and manager
**Describe the task**
As part of the move to a more scalable cloud architecture add a query service, which will be responsible for querying Sites and determining what should be replicated and to where, and placing those tasks on the queues.

kcover:
#### What does this PR do?
Adds the query service and manager. Creates the test-commons module which includes the embedded solr, and a junit rule for collecting errors during multithreaded testing. Creates component tests for the Query Service.

CI build

Fixes #225 

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/commons/src/main/java/com/connexta/replication/data/MetadataImpl.java
+++ b/commons/src/main/java/com/connexta/replication/data/MetadataImpl.java
@@ -51,17 +51,41 @@ public class MetadataImpl implements Metadata {
   private boolean isDeleted = false;
 
   /**
+   * Instantiates a metadata with the specified information.
+   *
    * @param metadata the raw metadata to wrap, cannot be null
    * @param type the type of the metadata, cannot be null
    * @param id id the of the metadata, cannot be null or empty
    * @param metadataModified the last time the metadata was modified, cannot be null
+   * @throws NullPointerException if any of the parameters are null
+   * @throws IllegalArgumentException if ID is empty or the type doesn't match the metadata
    */
   public MetadataImpl(Object metadata, Class type, String id, Date metadataModified) {
+    this(0L, metadata, type, id, metadataModified);
+  }
+
+  /**
+   * Instantiates a metadata with the specified information.
+   *
+   * @param metadataSize the size of the metadata
+   * @param metadata the raw metadata to wrap, cannot be null
+   * @param type the type of the metadata, cannot be null
+   * @param id id the of the metadata, cannot be null or empty
+   * @param metadataModified the last time the metadata was modified, cannot be null
+   * @throws NullPointerException if any of the parameters are null
+   * @throws IllegalArgumentException if ID is empty or the type doesn't match the metadata
+   */
+  public MetadataImpl(
+      long metadataSize, Object metadata, Class type, String id, Date metadataModified) {
+    if (!type.isInstance(metadata)) {
+      throw new IllegalArgumentException("Type must match the class of the metadata");
+    }
+
+    this.metadataSize = metadataSize;
     this.metadata = notNull(metadata, "Argument metadata may not be null");
     this.type = notNull(type, "Argument type may not be null");
     this.id = notEmpty(id, "Argument id may not be null or empty");
     this.metadataModified = notNull(metadataModified, "Argument metadataModified may not be null");
-
     this.tags = new HashSet<>();
     this.lineage = new ArrayList<>();
   }
@@ -156,5 +180,21 @@ public class MetadataImpl implements Metadata {
   @Override
   public void setIsDeleted(boolean isDeleted) {
     this.isDeleted = isDeleted;
+  }
+
+  /**
+   * Sets the resource info.
+   *
+   * @param size the size for the resource or <code>0</code> if no resource are available
+   * @param uri the uri for the resource or <code>null</code> if no resource is available
+   * @param modified the modified timestamp for the resource or <code>null</code> if no resource is
+   *     available
+   * @return this for chaining
+   */
+  public MetadataImpl setResource(long size, @Nullable URI uri, @Nullable Date modified) {
+    this.resourceSize = size;
+    this.resourceUri = uri;
+    this.resourceModified = modified;
+    return this;
   }
 }
--- a/query/pom.xml
+++ b/query/pom.xml
@@ -1,130 +1,29 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-<!--
-/**
- * Copyright (c) Connexta
- *
- * This is free software: you can redistribute it and/or modify it under the terms of the GNU Lesser General Public License as published by the Free Software Foundation, either
- * version 3 of the License, or any later version.
- *
- * This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
- * See the GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public License is distributed along with this program and can be found at
- * <http://www.gnu.org/licenses/lgpl.html>.
- *
- **/
--->
+<?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
          xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-    <modelVersion>4.0.0</modelVersion>
     <parent>
-        <groupId>com.connexta.replication</groupId>
         <artifactId>replication-service</artifactId>
+        <groupId>com.connexta.replication</groupId>
         <version>0.3.0-SNAPSHOT</version>
     </parent>
-    <artifactId>replication-api-impl</artifactId>
-    <name>Replication :: API Impl</name>
+    <modelVersion>4.0.0</modelVersion>
+
+    <name>Replication :: Query</name>
+    <artifactId>query</artifactId>
 
     <dependencies>
+        <dependency>
+            <groupId>com.connexta.replication</groupId>
+            <artifactId>replication-api-impl</artifactId>
+            <version>${project.version}</version>
+        </dependency>
         <dependency>
             <groupId>com.connexta.replication</groupId>
             <artifactId>replication-api</artifactId>
             <version>${project.version}</version>
         </dependency>
-        <dependency>
-            <groupId>com.connexta.replication</groupId>
-            <artifactId>commons</artifactId>
-            <version>${project.version}</version>
-        </dependency>
 
-        <dependency>
-            <groupId>org.springframework.data</groupId>
-            <artifactId>spring-data-solr</artifactId>
-        </dependency>
-        <!-- START spring-data-solr vulnerability workaround -->
-        <dependency>
-            <groupId>org.apache.solr</groupId>
-            <artifactId>solr-solrj</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-alpn-client</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-alpn-java-client</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-client</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-http</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-io</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty</groupId>
-            <artifactId>jetty-util</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty.http2</groupId>
-            <artifactId>http2-client</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty.http2</groupId>
-            <artifactId>http2-common</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty.http2</groupId>
-            <artifactId>http2-hpack</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.eclipse.jetty.http2</groupId>
-            <artifactId>http2-http-client-transport</artifactId>
-        </dependency>
-        <!-- END spring-data-solr vulnerability workaround -->
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.servicemix.bundles</groupId>
-            <artifactId>org.apache.servicemix.bundles.jsr305</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-collections4</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-lang3</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>net.jodah</groupId>
-            <artifactId>failsafe</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>io.micrometer</groupId>
-            <artifactId>micrometer-core</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>com.connexta.replication</groupId>
-            <artifactId>replication-solr-schema</artifactId>
-            <version>${project.version}</version>
-            <scope>test</scope>
-        </dependency>
         <dependency>
             <groupId>com.connexta.replication</groupId>
             <artifactId>test-commons</artifactId>
@@ -132,25 +31,23 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>commons-io</groupId>
-            <artifactId>commons-io</artifactId>
+            <groupId>org.springframework.data</groupId>
+            <artifactId>spring-data-solr</artifactId>
             <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>com.jparams</groupId>
-            <artifactId>to-string-verifier</artifactId>
-        </dependency>
-        <dependency>
-            <groupId>org.awaitility</groupId>
-            <artifactId>awaitility</artifactId>
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.apache.solr</groupId>
-            <artifactId>solr-core</artifactId>
+            <groupId>com.connexta.replication</groupId>
+            <artifactId>replication-solr-schema</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <version>${awaitility.version}</version>
         </dependency>
     </dependencies>
 
@@ -218,17 +115,17 @@
                                         <limit implementation="org.codice.jacoco.LenientLimit">
                                             <counter>INSTRUCTION</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.82</minimum>
+                                            <minimum>0.90</minimum>
                                         </limit>
                                         <limit implementation="org.codice.jacoco.LenientLimit">
                                             <counter>BRANCH</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.77</minimum>
+                                            <minimum>0.91</minimum>
                                         </limit>
                                         <limit implementation="org.codice.jacoco.LenientLimit">
                                             <counter>COMPLEXITY</counter>
                                             <value>COVEREDRATIO</value>
-                                            <minimum>0.80</minimum>
+                                            <minimum>0.86</minimum>
                                         </limit>
                                     </limits>
                                 </rule>
@@ -250,13 +147,6 @@
                     <plugin>
                         <groupId>org.owasp</groupId>
                         <artifactId>dependency-check-maven</artifactId>
-                        <configuration>
-                            <suppressionFiles>
-                                <suppressionFile>
-                                    ${basedir}/owasp-suppressions.xml
-                                </suppressionFile>
-                            </suppressionFiles>
-                        </configuration>
                     </plugin>
                 </plugins>
             </build>
--- a/query/src/main/java/com/connexta/replication/query/DdfMetadataInfoImpl.java
+++ b/query/src/main/java/com/connexta/replication/query/DdfMetadataInfoImpl.java
@@ -1,0 +1,104 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.query;
+
+import com.connexta.replication.api.data.Metadata;
+import com.connexta.replication.api.data.Task;
+import com.connexta.replication.api.data.ddf.DdfMetadataInfo;
+import java.time.Instant;
+import java.util.OptionalLong;
+
+/**
+ * Provides Information about metadata collected from a DDF, which is useful when processing a
+ * {@link Task}.
+ */
+public class DdfMetadataInfoImpl<T> implements DdfMetadataInfo<T> {
+
+  private String type;
+
+  private Instant lastModified;
+
+  private long size;
+
+  private Class<T> dataClass;
+
+  private T data;
+
+  /**
+   * Creates a new DdfMetadataInfoImpl by copying the necessary information from the given metadata.
+   *
+   * @param type A String describing the type of the metadata (e.g. DDMS 2.0, DDMS 5.0, ...)
+   * @param metadata The {@link Metadata} to create a DdfMetadataInfoImpl from
+   * @throws IllegalArgumentException if the type of the metadata does not match the type of the raw
+   *     metadata
+   */
+  public DdfMetadataInfoImpl(String type, Metadata metadata) {
+    this(
+        type,
+        metadata.getMetadataModified().toInstant(),
+        metadata.getMetadataSize(),
+        metadata.getType(),
+        (T) metadata.getRawMetadata());
+  }
+
+  /**
+   * Creates a new DdfMetadataInfoImpl with the provided information.
+   *
+   * @param type A String describing the type of the metadata (e.g. DDMS 2.0, DDMS 5.0, ...)
+   * @param lastModified The metadata modified time of the DDF metadata
+   * @param size The size of the metadata in bytes
+   * @param dataClass The class (i.e. type) of the raw metadata
+   * @param data The raw metadata
+   * @throws IllegalArgumentException if the type of the data does not match the given dataclass
+   */
+  public DdfMetadataInfoImpl(
+      String type, Instant lastModified, long size, Class<T> dataClass, T data) {
+    if (!dataClass.isInstance(data)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Metadata type was %s but the actual metadata was %s",
+              dataClass.getName(), data.getClass()));
+    }
+    this.type = type;
+    this.lastModified = lastModified;
+    this.size = size;
+    this.dataClass = dataClass;
+    this.data = data;
+  }
+
+  @Override
+  public String getType() {
+    return type;
+  }
+
+  @Override
+  public Instant getLastModified() {
+    return lastModified;
+  }
+
+  @Override
+  public OptionalLong getSize() {
+    return OptionalLong.of(size);
+  }
+
+  @Override
+  public Class<T> getDataClass() {
+    return dataClass;
+  }
+
+  @Override
+  public T getData() {
+    return data;
+  }
+}
--- a/query/src/main/java/com/connexta/replication/query/QueryManager.java
+++ b/query/src/main/java/com/connexta/replication/query/QueryManager.java
@@ -1,0 +1,205 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.query;
+
+import com.connexta.replication.api.AdapterException;
+import com.connexta.replication.api.AdapterInterruptedException;
+import com.connexta.replication.api.data.Site;
+import com.connexta.replication.api.data.SiteKind;
+import com.connexta.replication.api.queue.QueueException;
+import com.google.common.annotations.VisibleForTesting;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import javax.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/** Managers for instantiating {@link QueryService} class based on configuration. */
+public class QueryManager {
+  private static final Logger LOGGER = LoggerFactory.getLogger(QueryManager.class);
+
+  /** A Facade for the various objects and services used by the query service. */
+  private final QueryServiceTools queryServiceTools;
+
+  /** Set of identifiers for the sites to be managed or empty to manage all sites. */
+  private final Set<String> sites;
+
+  /** Executor service that controls the thread for regularly checking the configured sites. */
+  private final ScheduledExecutorService executor;
+
+  /** Map of active query services keyed by site IDs. */
+  private final Map<String, QueryService> services = new HashMap<>();
+
+  /**
+   * The number of seconds to wait before reloading the site configs and updating the {@link
+   * com.connexta.replication.query.QueryService}s accordingly.
+   */
+  private long serviceRefreshPeriod;
+
+  /**
+   * Instantiate a new query manager.
+   *
+   * @param sites identifiers for the sites to be managed or empty to manage all sites
+   * @param queryServiceTools facade for various objects used by the queryService
+   */
+  public QueryManager(
+      Stream<String> sites, QueryServiceTools queryServiceTools, long serviceRefreshPeriod) {
+    this(
+        sites,
+        queryServiceTools,
+        serviceRefreshPeriod,
+        Executors.newSingleThreadScheduledExecutor());
+  }
+
+  QueryManager(
+      Stream<String> sites,
+      QueryServiceTools queryServiceTools,
+      long serviceRefreshPeriod,
+      ScheduledExecutorService executor) {
+    this.sites = sites.collect(Collectors.toSet());
+    this.queryServiceTools = queryServiceTools;
+    this.serviceRefreshPeriod = serviceRefreshPeriod;
+    this.executor = executor;
+  }
+
+  /**
+   * Initializes the query manager and kick start all query services to start polling sites that
+   * must be polled.
+   */
+  public void init() {
+    if (sites.isEmpty()) {
+      LOGGER.info("Managing queries for all configured sites.");
+    } else {
+      LOGGER.info("Managing queries for sites: {}.", sites);
+    }
+    executor.scheduleAtFixedRate(
+        this::reloadSiteConfigs, 0, serviceRefreshPeriod, TimeUnit.SECONDS);
+  }
+
+  /** Destroys the query manager and stop all currently running query services. */
+  public void destroy() {
+    LOGGER.info("Shutting down query managers");
+    executor.shutdownNow();
+    synchronized (this) {
+      for (QueryService service : services.values()) {
+        service.stop();
+      }
+      services.clear();
+    }
+  }
+
+  @VisibleForTesting
+  void reloadSiteConfigs() {
+    Stream<Site> siteStream = queryServiceTools.sites();
+    synchronized (this) {
+      if (Thread.currentThread().isInterrupted()) {
+        return;
+      }
+
+      if (!sites.isEmpty()) {
+        siteStream = siteStream.filter(site -> sites.contains(site.getId()));
+      }
+      final Map<String, Site> newSites =
+          siteStream.collect(Collectors.toMap(Site::getId, Function.identity()));
+
+      // stop query services for sites that are no longer configured or have changed to a non-polled
+      // type
+      // update them if they are already running
+      // TODO: remove filters for regional DDF when we support other kinds of sites
+      try {
+        updateQueryServices(newSites);
+
+        // start query services for new sites that must be polled
+        newSites.values().stream()
+            .filter(QueryManager::mustBePolled)
+            .filter(site -> site.getKind() == SiteKind.REGIONAL)
+            .map(this::newService)
+            .filter(Objects::nonNull)
+            .forEach(QueryService::start);
+      } catch (AdapterInterruptedException e) {
+        LOGGER.debug(
+            "Thread interrupted while creating/updating QueryServices. This is normal when shutting down.",
+            e);
+        Thread.currentThread().interrupt();
+      }
+    }
+  }
+
+  @VisibleForTesting
+  void updateQueryServices(Map<String, Site> newSites) {
+    final Iterator<Entry<String, QueryService>> serviceIterator = services.entrySet().iterator();
+    while (serviceIterator.hasNext()) {
+      final Entry<String, QueryService> entry = serviceIterator.next();
+      final Site updatedSite = newSites.remove(entry.getKey());
+
+      if (updatedSite != null
+          && updatedSite.getType().mustBePolled()
+          && updatedSite.getKind() == SiteKind.REGIONAL) {
+        try {
+          entry.getValue().update(updatedSite);
+        } catch (AdapterInterruptedException e) {
+          throw e;
+        } catch (AdapterException | IllegalArgumentException e) {
+          LOGGER.warn(
+              "Failed to create an adapter for {} stopping the query service for now."
+                  + updatedSite);
+          LOGGER.debug("", e);
+          entry.getValue().stop();
+          serviceIterator.remove();
+        }
+      } else {
+        entry.getValue().stop();
+        serviceIterator.remove();
+      }
+    }
+  }
+
+  @VisibleForTesting
+  Map<String, QueryService> getServices() {
+    return new HashMap<>(services);
+  }
+
+  @VisibleForTesting
+  void putService(String id, QueryService service) {
+    services.put(id, service);
+  }
+
+  @Nullable
+  private QueryService newService(Site site) {
+    try {
+      return services.computeIfAbsent(site.getId(), i -> new QueryService(site, queryServiceTools));
+    } catch (AdapterInterruptedException e) {
+      throw e;
+    } catch (AdapterException | IllegalArgumentException | QueueException e) {
+      LOGGER.warn(
+          "Failed to create an adapter for {} not creating a query service for this site." + site);
+      LOGGER.debug("", e);
+      return null;
+    }
+  }
+
+  private static boolean mustBePolled(Site site) {
+    return site.getType().mustBePolled();
+  }
+}
--- a/query/src/main/java/com/connexta/replication/query/QueryService.java
+++ b/query/src/main/java/com/connexta/replication/query/QueryService.java
@@ -1,0 +1,283 @@
+/**
+ * Copyright (c) Connexta
+ *
+ * <p>This is free software: you can redistribute it and/or modify it under the terms of the GNU
+ * Lesser General Public License as published by the Free Software Foundation, either version 3 of
+ * the License, or any later version.
+ *
+ * <p>This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details. A copy of the GNU Lesser General Public
+ * License is distributed along with this program and can be found at
+ * <http://www.gnu.org/licenses/lgpl.html>.
+ */
+package com.connexta.replication.query;
+
+import com.connexta.replication.api.AdapterException;
+import com.connexta.replication.api.AdapterInterruptedException;
+import com.connexta.replication.api.NodeAdapter;
+import com.connexta.replication.api.ReplicationException;
+import com.connexta.replication.api.data.Filter;
+import com.connexta.replication.api.data.FilterIndex;
+import com.connexta.replication.api.data.Metadata;
+import com.connexta.replication.api.data.MetadataInfo;
+import com.connexta.replication.api.data.OperationType;
+import com.connexta.replication.api.data.QueryRequest;
+import com.connexta.replication.api.data.Site;
+import com.connexta.replication.api.data.TaskInfo;
+import com.connexta.replication.api.queue.QueueException;
+import com.connexta.replication.api.queue.SiteQueue;
+import com.connexta.replication.data.QueryRequestImpl;
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.Date;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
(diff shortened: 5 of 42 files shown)


query/src/main/java/com/connexta/replication/query/DdfMetadataInfoImpl.java
```diff
@@ -0,0 +1,74 @@
...
+
+public class DdfMetadataInfoImpl implements DdfMetadataInfo {
+
+  private String type;
+
+  private Instant lastModified;
+
+  private long size;
+
+  private Class<?> dataClass;
```
paouelle: `private final Class<T> dataClass;`

query/src/main/java/com/connexta/replication/query/DdfMetadataInfoImpl.java
```diff
@@ -0,0 +1,74 @@
...
+
+  private Object data;
```
paouelle: `private final T data;`

query/src/main/java/com/connexta/replication/query/DdfMetadataInfoImpl.java
```diff
@@ -0,0 +1,74 @@
...
+    return lastModified;
+  }
+
+  @Override
+  public OptionalLong getSize() {
+    return OptionalLong.of(size);
+  }
+
+  @Override
+  public Class<?> getDataClass() {
```
paouelle: `public Class<T> getDataClass()`

query/src/main/java/com/connexta/replication/query/DdfMetadataInfoImpl.java
```diff
@@ -0,0 +1,74 @@
...
+    return dataClass;
+  }
+
+  @Override
+  public Object getData() {
```
paouelle: `public T getData()`

query/src/main/java/com/connexta/replication/query/QueryServiceTools.java
```diff
@@ -0,0 +1,89 @@
...
+import com.connexta.replication.api.impl.NodeAdapters;
+import com.connexta.replication.api.persistence.FilterIndexManager;
+import com.connexta.replication.api.persistence.FilterManager;
+import com.connexta.replication.api.persistence.SiteManager;
+import com.connexta.replication.api.queue.QueueBroker;
+import com.connexta.replication.api.queue.SiteQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
+
+public class QueryServiceTools {
```
paouelle: ❗️ javadoc

query/src/main/java/com/connexta/replication/query/QueryServiceTools.java
```diff
@@ -0,0 +1,89 @@
...
+  private final FilterIndexManager filterIndexManager;
+
+  private final QueueBroker queueBroker;
+
+  private final long period;
+
+  /** Default polling period for configuration changes in seconds. */
+  private static final long DEFAULT_REPLICATION_PERIOD = TimeUnit.MINUTES.toSeconds(5);
+
+  public QueryServiceTools(
```
paouelle: ✏️ javadoc

query/src/main/java/com/connexta/replication/query/QueryServiceTools.java
```diff
@@ -0,0 +1,89 @@
...
+      long period) {
+    this.nodeAdapterFactories = nodeAdapterFactories;
+    this.siteManager = siteManager;
+    this.filterManager = filterManager;
+    this.filterIndexManager = filterIndexManager;
+    this.queueBroker = queueBroker;
+    this.period = period > 0L ? period : DEFAULT_REPLICATION_PERIOD;
+  }
+
+  public long getPeriod() {
```
paouelle: ✏️ javadoc all public methods

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+  public QueryService(Site site, QueryServiceTools queryServiceTools) {
+    this(site, queryServiceTools, Executors.newSingleThreadScheduledExecutor());
+  }
+
+  public QueryService(
+      Site site, QueryServiceTools queryServiceTools, ScheduledExecutorService executor) {
+    this.queryServiceTools = queryServiceTools;
+    this.siteQueue = queryServiceTools.getQueueFor(site);
+    this.executor = executor;
+    update(site);
```
paouelle: ❗️ Should not be calling public methods from a constructor unless the class is final

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+  /**
+   * Instantiates a new query service.
+   *
+   * @param site the configuration for the site to be queried by this query service
+   */
+  public QueryService(Site site, QueryServiceTools queryServiceTools) {
+    this(site, queryServiceTools, Executors.newSingleThreadScheduledExecutor());
+  }
+
+  public QueryService(
```
paouelle: ✏️ annotate with `@Nullable` and remove `public`

query/src/test/java/com/connexta/replication/query/QueryServiceToolsTest.java
```diff
@@ -0,0 +1,128 @@
...
+            period);
+  }
+
+  @Test
+  public void getPeriod() {
+    assertThat(queryServiceTools.getPeriod(), Matchers.is(period));
+  }
+
+  @Test
+  public void getPeriodDefault() {
```
paouelle: ✏️ This is more testing the constructor as opposed to the getter.

query/src/test/java/com/connexta/replication/query/QueryServiceToolsTest.java
```diff
@@ -0,0 +1,128 @@
...
+    FilterIndex filterIndex = mock(FilterIndex.class);
+    queryServiceTools.saveFilterIndex(filterIndex);
+    verify(filterIndexManager).save(filterIndex);
+  }
+
+  @Test
+  public void getQueueFor() {
+    Site site = mock(Site.class);
+    when(site.getId()).thenReturn("siteid");
+    queryServiceTools.getQueueFor(site);
```
paouelle: ❗️ There is no verification of the result here or the fact the queue broker is called with the id

query/src/test/java/com/connexta/replication/query/QueryServiceToolsTest.java
```diff
@@ -0,0 +1,128 @@
...
+    queryServiceTools.getAdapterFor(site);
+    verify(nodeAdapterFactory).create(null);
+  }
+
+  @Test
+  public void activeFiltersFor() {
+    Filter filter = mock(Filter.class);
+    Site site = mock(Site.class);
+    when(filterManager.filtersForSite(null)).thenReturn(Stream.of(filter));
+    assertThat(queryServiceTools.activeFiltersFor(site).findFirst().get(), Matchers.is(filter));
```
paouelle: ✏️ Although yu are verifying you are getting `filter` back, you are not verifying that this is the only thing you are getting back. I would suggest you collect the result into a list and then use `Matchers.contains()` or `Matchers.containsInAnyorder()` to verify the content is exactly what you expect and nothing more

query/src/main/java/com/connexta/replication/query/QueryManager.java
```diff
@@ -0,0 +1,128 @@
...
+        TimeUnit.SECONDS);
+  }
+
+  /** Destroys the query manager and stop all currently running query services. */
+  public void destroy() {
+    LOGGER.info("Shutting down query managers");
+    executor.shutdownNow();
+    for (QueryService service : services.values()) {
+      service.stop();
+    }
```
paouelle: ✏️ We should probably clear the `services` map.

query/src/main/java/com/connexta/replication/query/QueryManager.java
```diff
@@ -0,0 +1,128 @@
...
+
+  void updateQueryServices(Map<String, Site> newSites) {
+    for (Entry<String, QueryService> entry : services.entrySet()) {
+      final Site updatedSite = newSites.remove(entry.getKey());
+
+      if (updatedSite != null && updatedSite.getType().mustBePolled()) {
+        entry.getValue().update(updatedSite);
+      } else {
+        services.remove(entry.getKey());
+        entry.getValue().stop();
```
paouelle: ❗️ All sites that are stopped should also be removed from `services`

query/src/main/java/com/connexta/replication/query/QueryManager.java
```diff
@@ -0,0 +1,128 @@
...
+        queryServiceTools.allSites().collect(Collectors.toMap(Site::getId, Function.identity()));
+
+    // stop query services for sites that are no longer configured or have changed to a non-polled
+    // type
+    // update them if they are already running
+    updateQueryServices(newSites);
+
+    // start query services for new sites that must be polled
+    newSites.values().stream()
+        .filter(QueryManager::mustBePolled)
```
paouelle: ❗️ We should also filter for only the sites that are in our `sites` list if that one is not empty. Either here or at the line above where we get all sites from config.

query/src/test/java/com/connexta/replication/query/QueryManagerTest.java
```diff
@@ -0,0 +1,121 @@
...
+  private QueryManager queryManager;
+
+  @Before
+  public void setUp() throws Exception {
+    queryManager = new QueryManager(sites, queryServiceTools, executor);
+    when(queryServiceTools.getPeriod()).thenReturn(PERIOD);
+    when(site.getId()).thenReturn(SITE_ID);
+    when(site.getType()).thenReturn(SiteType.DDF);
+  }
+
```
paouelle: ✏️ We should test cases where no site ids are defined (what we have here) and some when site ids are defined.

query/src/test/java/com/connexta/replication/query/QueryManagerTest.java
```diff
@@ -0,0 +1,121 @@
...
+  @Mock QueryServiceTools queryServiceTools;
+
+  @Mock ScheduledExecutorService executor;
+
+  @Mock private Site site;
+
+  @Mock private QueryService queryService;
+
+  private Stream<String> sites = Stream.empty();
```
paouelle: ❗️ Do not save `Stream` as attribute or local variables. It should be avoided.

query/src/test/java/com/connexta/replication/query/QueryManagerTest.java
```diff
@@ -0,0 +1,121 @@
...
+    verify(queryService).stop();
+  }
+
+  @Test
+  public void reloadSiteConfigs() {
+    when(queryServiceTools.allSites()).thenReturn(Stream.of(site));
+    when(queryServiceTools.getQueueFor(site)).thenReturn(mock(SiteQueue.class));
+    queryManager.reloadSiteConfigs();
+    // assert that a query service was started
+    assertThat(queryManager.services.get(SITE_ID).future, Matchers.notNullValue());
```
paouelle: ❗️ Avoid accessing private members of another class to verify something. Instead, expose a package private method to check its state that hides how to do so. Something like `isStarted()` maybe

paouelle: ❓ Should we also verify that the service is being tracked internally now?

query/src/test/java/com/connexta/replication/query/QueryManagerTest.java
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  @Test
+  public void updateQueryServicesRemovedSite() {
+    queryManager.services.put(SITE_ID, queryService);
+    Map<String, Site> newSites = mock(Map.class);
+    queryManager.updateQueryServices(newSites);
+    assertThat(queryManager.services.keySet(), Matchers.not(Matchers.contains(SITE_ID)));
+    verify(queryService).stop();
```
paouelle: ✏️ We should verify that the site is no longer being tracked by the manager.

query/src/test/java/com/connexta/replication/query/QueryManagerTest.java
paouelle: ✏️ We might want to add a test case that will:
- add a new site to manage
- remove an existing site to manage
- add a new site that would be filtered

Just to make sure that we do handle the combination

query/src/test/java/com/connexta/replication/query/QueryManagerTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+  @Test
+  public void updateQueryServicesNoLongerPolledSite() {
+    queryManager.services.put(SITE_ID, queryService);
+    Map<String, Site> newSites = mock(Map.class);
+    Site notPolledSite = mock(Site.class);
+    when(notPolledSite.getType()).thenReturn(SiteType.ION);
+    when(newSites.remove(SITE_ID)).thenReturn(notPolledSite);
+    queryManager.updateQueryServices(newSites);
+    verify(queryService).stop();
```
paouelle: ✏️ We should verify that the site is no longer being tracked by the manager.

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+  public void update(Site site) {
+    synchronized (this) {
+      this.site = site;
+      try {
+        this.adapter = queryServiceTools.getAdapterFor(site);
+      } catch (AdapterException e) {
+        LOGGER.info(
+            "Failed to create an adapter for {} stopping the query service for now."
+                + site.getUrl());
+        stop();
```
paouelle: ✏️ The problem I see here is that we are stopping the site but it will never get removed from the manager. Should we instead let the exception bubble out back to the manager and let the manager handle stopping the service and removing it from its set of managed services?

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+      }
+    }
+  }
+
+  /** Starts this query service. */
+  public void start() {
+    if (future == null || future.isDone()) {
+      future =
```
paouelle: ✏️ `this.future`

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+          executor.scheduleAtFixedRate(
+              this::query, 0, queryServiceTools.getPeriod(), TimeUnit.SECONDS);
+    }
+  }
+
+  /** Stops this query service. */
+  public void stop() {
+    executor.shutdownNow();
```
paouelle: ❓ Should we also cancel the future?

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+      return;
+    }
+
+    Set<Filter> filters = queryServiceTools.activeFiltersFor(site).collect(Collectors.toSet());
+    for (Filter filter : filters) {
+      FilterIndex filterIndex = queryServiceTools.getOrCreateFilterIndex(filter);
+
+      // holds the modified time of each task put in the queue for this filter
+      // this will be used to grab the latest modified time after
+      Set<Instant> modifiedTimes = new HashSet<>();
```
paouelle: ❓ Why not keep the latest one and update it if you find one that is later than that?

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+  }
+
+  @VisibleForTesting
+  void query() {
+    if (!adapter.isAvailable()) {
+      LOGGER.debug("System at " + site.getUrl() + " is currently unavailable");
+      return;
+    }
+
+    Set<Filter> filters = queryServiceTools.activeFiltersFor(site).collect(Collectors.toSet());
```
paouelle: ✏️ We should probably sort the filters by priority such that we do the highest one first.

paouelle: ✏️ Make it `final`

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+    for (Filter filter : filters) {
+      FilterIndex filterIndex = queryServiceTools.getOrCreateFilterIndex(filter);
```
paouelle: ✏️ Make it `final`

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+            Collections.emptyList(), // ignoring excluded nodes (origins tag)
+            Collections.emptyList(), // ignoring failed items
+            indexDate);
+
+    Iterable<Metadata> changeSet = adapter.query(queryRequest).getMetadata();
+    return StreamSupport.stream(changeSet.spliterator(), false);
+  }
+
+  private TaskInfo metadataToTask(Metadata metadata, byte filterPriority) {
+    TaskInfoImpl task = new TaskInfoImpl();
```
paouelle: ✏️ Make it `final`

query/src/main/java/com/connexta/replication/query/TaskInfoImpl.java
```diff
@@ -0,0 +1,137 @@
...
+
+  public void setLastModified(Instant lastModified) {
+    this.lastModified = lastModified;
+  }
+
+  public void setResourceInfo(@Nullable ResourceInfo resourceInfo) {
+    this.resourceInfo = resourceInfo;
+  }
+
+  public void setMetadatas(Stream<MetadataInfo> metadatas) {
```
paouelle: ✏️ Instead of set, how about `add` methods where you can pass in a stream of metadata, `MetadataInfo...`. This would make it easier on your clients and allow you to collect them internally on a list.

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+    // TODO: determine the operation type based on the site type and kind
+    return OperationType.HARVEST;
+  }
+
+  private void queueTaskAndPutLastModifiedInSet(TaskInfo task, Set<Instant> modifiedTimes) {
+    try {
+      siteQueue.put(task);
+    } catch (InterruptedException e) {
+      modifiedTimes.add(task.getLastModified());
+      throw new QueueException(
```
paouelle: ❗️ I would suggest you let that exception bubble out as opposed to convert it into an error. This is an indication that you are stopping this service based on the design. Let it bubble out, and above, handle it by updating your index if needed, persist it, and then bubble it out again (or reset the interrupted flag for the thread) before exiting the `query()` method

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+
+    // not implementing failure retries here
+    QueryRequest queryRequest =
+        new QueryRequestImpl(
+            filter.getFilter(),
+            Collections.emptyList(), // ignoring excluded nodes (origins tag)
+            Collections.emptyList(), // ignoring failed items
+            indexDate);
+
+    Iterable<Metadata> changeSet = adapter.query(queryRequest).getMetadata();
```
paouelle: ❗️ ❓ How does the adapter deals with `InterrutedIOException`? Does it blindly convert it into an `AdapterException`? If so then we need to update the adapter to handle it differently as you need to be made aware that you got interrupted other than by receiving a generic error since being interrupted is not an error.

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+  private Stream<Metadata> getNewMetadataForFilter(Filter filter, FilterIndex filterIndex)
+      throws AdapterException {
+    // get last modified date
+    final Date indexDate = filterIndex.getModifiedSince().map(Date::from).orElse(null);
+    if (indexDate == null) {
+      LOGGER.trace("No index for filter {} found.", filter.getName());
+    }
+
+    // not implementing failure retries here
+    QueryRequest queryRequest =
```
paouelle: ✏️ Make it final

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+      } catch (AdapterException e) {
+        LOGGER.debug(
+            "Couldn't query site {}. Stopping queries for this site until next polling attempt.",
+            site.getName(),
+            e);
+        return;
+      } catch (QueueException qe) {
+        LOGGER.debug(
+            "Failed to put a task in queue for site {}. Stopping queuing of tasks for this site. Will resume at next polling attempt.",
+            site.getName(),
```
paouelle: ❓ Should we still update the index for what we have do so far?

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,219 @@
...
+
+    return task;
+  }
+
+  private OperationType determineOperationType() {
+    // TODO: determine the operation type based on the site type and kind
+    return OperationType.HARVEST;
+  }
+
+  private void queueTaskAndPutLastModifiedInSet(TaskInfo task, Set<Instant> modifiedTimes) {
```
paouelle: ✏️ Consider to simply return the task created such that you can do a `map()` instead of a `forEach()` and leave it up to the loop to update the latest modified time as opposed to scattering this logic across multiple methods.

query/src/test/java/com/connexta/replication/query/QueryServiceTest.java
```diff
@@ -0,0 +1,165 @@
...
+import java.util.stream.Stream;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class QueryServiceTest {
```
paouelle: ❗️ We should definitely create test cases that verifies the behavior of being stopped by the scheduler via interruptions when we are querying, and when we are creating and queuing tasks to make sure we keep our index up to date.

query/src/test/java/com/connexta/replication/query/QueryServiceTest.java
```diff
@@ -0,0 +1,165 @@
...
+            any(Runnable.class), eq(0L), eq(PERIOD), eq(TimeUnit.SECONDS)))
+        .thenReturn(future);
+    queryService.start();
+    verify(executor)
+        .scheduleAtFixedRate(any(Runnable.class), eq(0L), eq(PERIOD), eq(TimeUnit.SECONDS));
+    assertThat(queryService.future, notNullValue());
+  }
+
+  @Test
+  public void query() throws Exception {
```
paouelle: ✏️ You need test cases that will ensure that you do keep the right modified timestamp when dealing with multiple metacards in a single query and a mixture of them.

query/src/test/java/com/connexta/replication/query/QueryServiceTest.java
```diff
@@ -0,0 +1,165 @@
...
+    when(filter.getPriority()).thenReturn((byte) 0);
+    when(metadata.getResourceSize()).thenReturn(0L);
+    Date lastModified = new Date();
+    when(metadata.getMetadataModified()).thenReturn(lastModified);
+
+    queryService.query();
+
+    verify(adapter).query(requestArgumentCaptor.capture());
+    assertThat(requestArgumentCaptor.getValue().getModifiedAfter(), is(nullValue()));
+    verify(siteQueue).put(any(TaskInfoImpl.class));
```
paouelle: ❓ How do we know if the task created is the right one? Contains the right resource info and metadata info, the right priority, ...
We should have test cases that tests a few combination of those.

query/pom.xml
```diff
@@ -0,0 +1,29 @@
...
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <name>Replication :: Query</name>
+    <artifactId>query</artifactId>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.connexta.replication</groupId>
+            <artifactId>replication-api-impl</artifactId>
```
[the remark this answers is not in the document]
kcover: story added.

query/src/main/java/com/connexta/replication/query/DdfMetadataInfoImpl.java
```diff
@@ -0,0 +1,74 @@
...
+  public DdfMetadataInfoImpl(String type, Metadata metadata) {
+    this.type = type;
+    this.lastModified = metadata.getMetadataModified().toInstant();
+    this.size = metadata.getMetadataSize();
+    this.dataClass = metadata.getType();
+    this.data = metadata.getRawMetadata();
+  }
+
+  public DdfMetadataInfoImpl(
+      String type, Instant lastModified, long size, Class<?> dataClass, Object data) {
```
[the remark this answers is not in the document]
kcover: Shouldn't the type parameter go in the class declaration?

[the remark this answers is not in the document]
paouelle: Definitely better yes. That way, the other methods on this class would return `T` and accept `T`. See the update I did to `DDFMetadataInfo` in my PR

query/src/main/java/com/connexta/replication/query/QueryManager.java
kcover: That's what's happening in the line above but this comment tells me the code isn't very readable. Perhaps just switching the order of the lines would help?

paouelle: I did miss that. I would be careful about removing from a collection you are iterating without going through an actual iterator. This can tend to lead to concurrent modification exceptions. The best approach is to get an iterator on the entry set and iterate using that. When you want to remove the entry, you simply call `i.remove()`. This is more optimal since the iterator is already tracking where it is and avoid the concurrent modification exception that can occur.

query/src/main/java/com/connexta/replication/query/QueryService.java
kcover: What part is nullable?

query/src/main/java/com/connexta/replication/query/QueryService.java
kcover: I guess I'll just make the class final. Or I think I could just make the method final.

kcover: Nvm I just learned that you can't mock/spy a final class and since we are going to let the adapter exception bubble out I think I'll just set the fields directly.

query/src/main/java/com/connexta/replication/query/QueryService.java
kcover: Shutting down the executor will cancel the future as well but I will cancel the future first here for clarity.

query/src/main/java/com/connexta/replication/query/QueryService.java
paouelle: oh man. At this point, my mind is the only thing I can think of 😄

query/src/main/java/com/connexta/replication/query/QueryService.java
kcover: Yes. good catch

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,223 @@
...
+            filter.getFilter(),
+            Collections.emptyList(), // ignoring excluded nodes (origins tag)
+            Collections.emptyList(), // ignoring failed items
+            indexDate);
+
+    Iterable<Metadata> changeSet = adapter.query(queryRequest).getMetadata();
+    return StreamSupport.stream(changeSet.spliterator(), false);
+  }
+
+  private TaskInfo metadataToTask(Metadata metadata, byte filterPriority) {
```
paouelle: ❗️ One thing that will need to be enhanced later here is the fact that for regional DDF sites, we don't have to handle deleted pieces of intel since we are only harvesting from those whereas for other combinations, we do have to handle deletes and this means we will need to figure out a way to get that info and propagate delete operations.

For now, we should ensure to filter out any sites that are not regional DDF from the list of sites we start query services for inside the manager.

query/src/main/java/com/connexta/replication/query/QueryService.java
kcover: Not sure what you mean here. The task is created in another method and how would the loop update the latest modified time?

query/src/main/java/com/connexta/replication/query/TaskInfoImpl.java
```diff
@@ -0,0 +1,137 @@
...
+import java.time.Instant;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Stream;
+import javax.annotation.Nullable;
+
+public class TaskInfoImpl implements TaskInfo {
+
+  private String id;
```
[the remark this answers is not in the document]
kcover: Removed the setters and made the constructors require all the necessary fields.

query/src/main/java/com/connexta/replication/query/TaskInfoImpl.java
```diff
@@ -0,0 +1,137 @@
...
+
+  private byte priority;
+
+  private OperationType operation;
+
+  private Instant lastModified;
+
+  @Nullable private ResourceInfo resourceInfo;
+
+  private Stream<MetadataInfo> metadatas;
```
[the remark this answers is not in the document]
kcover: Using a set instead.

query/src/main/java/com/connexta/replication/query/TaskInfoImpl.java
```diff
@@ -0,0 +1,137 @@
...
+  public Optional<ResourceInfo> getResource() {
+    return Optional.of(resourceInfo);
+  }
+
+  @Override
+  public Stream<MetadataInfo> metadatas() {
+    return metadatas;
+  }
+
+  public void setId(String id) {
```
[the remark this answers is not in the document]
kcover: I removed the setters but why again do we document getters and setters that aren't doing special?

query/src/main/java/com/connexta/replication/query/QueryService.java
paouelle: I think we talked about that yesterday. What I proposed cannot be done here since we want to keep track of the latest modified timestamp as we iterate the stream otherwise when the stream is interrupted, we will loose that info.

query/src/test/java/com/connexta/replication/query/QueryManagerTest.java
kcover: tracked internally?

paouelle: Meaning that the manager actually is managing that service such that when it shuts down, it will shut it down for example?

paouelle: BTW, the manager needs to define a destroy() method so it can be registered with spring to be told to shutdown

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,223 @@
...
+  /**
+   * Updates the configuration for the site currently being queried by this query service.
+   *
+   * @param site the new site configuration
+   */
+  public void update(Site site) {
+    synchronized (this) {
+      this.site = site;
+      try {
+        this.adapter = queryServiceTools.getAdapterFor(site);
```
paouelle: ❗️ `getAdapterFor()` can throw `IllegalArgumentException` if it is unable to find a factory for the provided site.

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,223 @@
...
+      }
+    }
+  }
+
+  /** Starts this query service. */
+  public void start() {
+    if (!isRunning()) {
+      future =
+          executor.scheduleAtFixedRate(
+              this::query, 0, queryServiceTools.getPeriod(), TimeUnit.SECONDS);
```
paouelle: ❗️ We should be using the polling period that is configured with the site if any. If none, then we should be using a system one. However, I am not sure we should be using the same period we are using for polling for site config changes.

query/src/test/java/com/connexta/replication/query/QueryManagerTest.java
kcover: this is checking if the service was started by getting it from the managers list of services. I'm not sure how else we would check if they're tracked internally.

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+        List.of(
+            generateMetadataWithResource(111),
+            generateMetadataWithResource(112),
+            generateMetadataWithResource(113, new Date(1)));
+    when(adapter.query(Mockito.argThat(isQueryRequestWithFilter(filter.getFilter()))))
+        .thenReturn(
+            new QueryResponseImpl(
+                metadataList,
+                e -> {
+                  throw new AssertionError(e);
```
paouelle: ❗️ Make sure to record this with the collector or there is no chance it will ever be reported to fail the test

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+                }));
+
+    indexManager = Mockito.spy(indexManager);
```
paouelle: ❓ Wouldn't you need to pass the spied index manager to your tools in order for it to use the spy and you being able to verify calls on it?

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+    List<Metadata> metadataList =
+        List.of(
+            generateMetadataWithResource(111),
+            generateMetadataWithResource(112),
+            generateMetadataWithResource(113, new Date(1)));
+    when(adapter.query(Mockito.argThat(isQueryRequestWithFilter(filter.getFilter()))))
+        .thenReturn(
+            new QueryResponseImpl(
+                metadataList,
+                e -> {
```
paouelle: ✏️ Can be simply replace with `AssertionException::new`

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+          Throwable t = new AdapterException("no adapters for ION sites should be retrieved");
+          collector.addError(t);
+          throw t;
+        };
+  }
+
+  @After
+  public void cleanup() throws Exception {
+    tasks.clear();
+    filterManager.objects().map(Filter::getId).forEach(filterManager::remove);
```
paouelle: ✏️ It might be worthwhile (in a different ticket) to enhance our embedded solr server such that we can easily ask it to clear the DB which internally would blindly delete the content of all Solr tables. This is much simpler and much more generic than having to rely on our own code to query and delete everything. It is also important that cleaning up aborts testing if it fails to do so otherwise we will end up with leaks in between tests.

We might even be able to autowire that embedded server (or let it be defined here) allowing us to annotate it with `@Rule` such that we can detect the end of a test case and cleanup automatically no matter what.

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+    for (Metadata metadata : metadataList) {
+      assertTask(queue.take(), metadata, (byte) 0, OperationType.HARVEST);
+    }
+    verify(indexManager)
+        .save(
+            Mockito.argThat(
+                isIndexWithModified(metadataList.get(2).getMetadataModified().toInstant())));
+
+    for (final Task task : tasks) {
+      task.complete();
```
paouelle: ❗️ There is a danger here of having a test that sometime will work and sometime might fail. This is because of the instability created by the first `task.complete()`. When the first one is called, the queue will automatically wakeup the thread waiting to push the 4th task in it. If that task has higher priority, it might end up being pulled as the second task instead of being the 4th. We need to ensure that this doesn't happen in a consistent way. We can talk about it later.

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+                }),
+            new QueryResponseImpl(
+                List.of(secondPollMetadata),
+                e -> {
+                  throw new AssertionError(e);
+                }));
+
+    indexManager = Mockito.spy(indexManager);
+
+    startQueryManager(THIRTY_MINUTES);
```
paouelle: ❓ Wouldn't this make this test take 30 minutes to run? We need to scale time to 1 ms (i.e. the minimum) and introduce ways to ensure the test always run the same way.

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+
+    assertThat(
+        "timed out waiting for expected published task",
+        queue.waitForPendingSizeToReach(3, 5L, TimeUnit.SECONDS),
+        Matchers.equalTo(true));
+
+    for (Metadata metadata : metadataList) {
+      assertTask(queue.take(), metadata, (byte) 0, OperationType.HARVEST);
+    }
+    verify(indexManager)
```
paouelle: ✏️ How about we verify the index before we pull out tasks since we know te query service should have completed its pass and updated it. I would suggest we do that on all of the tests.

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+    startQueryManager(THIRTY_MINUTES);
+
+    final MemorySiteQueue queue = (MemorySiteQueue) broker.getQueue(site.getId());
+
+    assertThat(
+        "timed out waiting for expected published task",
+        queue.waitForPendingSizeToReach(2, 5L, TimeUnit.SECONDS),
+        Matchers.equalTo(true));
+
+    assertTask(queue.take(), metadataList.get(2), (byte) 9, OperationType.HARVEST);
```
paouelle: ✏️ We should verify the indexes were updated

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+    // them.
+    MemorySiteQueue tacticalQueue = (MemorySiteQueue) broker.getQueue(siteList.get(0).getId());
+    MemorySiteQueue ionQueue = (MemorySiteQueue) broker.getQueue(siteList.get(1).getId());
+    MemorySiteQueue regionalQueue = (MemorySiteQueue) broker.getQueue(siteList.get(2).getId());
+
+    assertThat(
+        "timed out waiting for expected published task",
+        regionalQueue.waitForPendingSizeToReach(1, 5L, TimeUnit.SECONDS),
+        Matchers.equalTo(true));
+    assertTask(regionalQueue.take(), metadataList.get(2), (byte) 0, OperationType.HARVEST);
```
paouelle: ✏️ We should verify all 3 indexes as well.

query/src/test/java/com/connexta/replication/query/QueryServiceTest.java
```diff
@@ -0,0 +1,574 @@
...
+    when(queryServiceTools.activeFiltersFor(site)).thenReturn(Stream.of(filter));
+    FilterIndex filterIndex = mock(FilterIndex.class);
+    when(queryServiceTools.getOrCreateFilterIndex(filter)).thenReturn(filterIndex);
+
+    // #getNewMetadataForFilter
+    when(filterIndex.getModifiedSince()).thenReturn(Optional.empty());
+    when(adapter.query(any(QueryRequest.class)))
+        .thenThrow(new AdapterInterruptedException(new InterruptedException()));
+
+    queryService.query();
```
paouelle: ❗️ We should verify that the thread state is interrupted at exit from this call. And because of the nature of these tests, make sure you use the `ClearInterruptions` JUnit rule from codice-test so that the thread state is properly cleared at the end of these tests.

BTW, this is not testing that query throws the exception but more that it handles it.

query/src/test/java/com/connexta/replication/query/QueryServiceTest.java
```diff
@@ -0,0 +1,574 @@
...
+
+    verify(siteQueue, never()).put(any(TaskInfoImpl.class));
+    verify(filterIndex).setModifiedSince(Instant.MIN);
+    verify(queryServiceTools).saveFilterIndex(filterIndex);
+  }
+
+  @Test
+  public void queryThrowsInterruptedException() throws Exception {
```
paouelle: ✏️ Maybe be clearer about the fact that you are testing interruptions when putting tasks in the queue

query/src/test/java/com/connexta/replication/query/QueryServiceTest.java
paouelle: ✏️ Might want to add one when the interruptions occurs while iterating the stream of metacards returned.

query/src/main/java/com/connexta/replication/query/QueryServiceTools.java
```diff
@@ -0,0 +1,146 @@
...
+   * @param queueBroker A service used to retrieve Queues
+   * @param period The length between polling intervals in seconds.
+   */
+  public QueryServiceTools(
+      NodeAdapters nodeAdapterFactories,
+      SiteManager siteManager,
+      FilterManager filterManager,
+      FilterIndexManager filterIndexManager,
+      QueueBroker queueBroker,
+      long period) {
```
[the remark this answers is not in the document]
kcover: Not sure I understand what you're asking but we now have another application property that will set the period between query service refresh. That way there is a difference between the QueryService refresh period and the period at which each query service queries its site.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/persistence/spring/FilterRepositoryIntegrationTest.java
```diff
@@ -23,27 +23,24 @@
...
-import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.data.solr.UncategorizedSolrException;
 import org.springframework.data.solr.repository.config.EnableSolrRepositories;
-import org.springframework.data.solr.server.SolrClientFactory;
 import org.springframework.test.context.junit4.SpringRunner;
 
 @ComponentScan(
-    basePackageClasses = FilterRepository.class) // to make sure we pickup the repositories
+    basePackageClasses = {EmbeddedSolrConfig.class}) // to make sure we pickup the repositories
```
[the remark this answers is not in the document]
kcover: It still works so I assume the annotation below this one is what picks up those repos.

query/src/main/java/com/connexta/replication/query/spring/ServiceConfig.java
```diff
@@ -0,0 +1,72 @@
...
+            nodeAdapters,
+            siteManager,
+            filterManager,
+            filterIndexManager,
+            queueBroker,
+            replicationProperties.getPeriod());
+    return new QueryManager(
+        replicationProperties.getSites().stream(),
+        queryServiceTools,
+        queryProperties.getServiceRefreshPeriod());
```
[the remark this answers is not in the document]
kcover: Only the QueryManager cares about this period since it's for checking the site configurations. It doesn't need to be passed around with the tools.

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,278 @@
...
+   */
+  @VisibleForTesting
+  void query() {
+    if (!adapter.isAvailable()) {
+      LOGGER.debug("System at " + site.getUrl() + " is currently unavailable");
+      return;
+    }
+
+    final List<Filter> filters =
+        queryServiceTools
```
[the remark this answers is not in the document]
kcover: It can? It's using the spring repo, not either of the adapters.

query/src/main/java/com/connexta/replication/query/QueryManager.java
```diff
@@ -0,0 +1,199 @@
...
+    while (serviceIterator.hasNext()) {
+      final Entry<String, QueryService> entry = serviceIterator.next();
+      final Site updatedSite = newSites.remove(entry.getKey());
+
+      if (updatedSite != null
+          && updatedSite.getType().mustBePolled()
+          && updatedSite.getKind() == SiteKind.REGIONAL) {
+        try {
+          entry.getValue().update(updatedSite);
+        } catch (AdapterException | IllegalArgumentException e) {
```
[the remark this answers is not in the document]
kcover: All we're doing is creating an adapter here. Can that throw an AdapterInterruptedException?

query/src/main/java/com/connexta/replication/query/QueryServiceTools.java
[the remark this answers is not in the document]
paouelle: To me, I see we have 2 configuration parameters:
- config polling period (what you have in the tools right now)
- default site polling period (what you currently pass out of band to the manager)

I am suggesting that all config be managed via your tools class as opposed to have some of the config out of band like it is today.

query/src/main/java/com/connexta/replication/query/spring/ServiceConfig.java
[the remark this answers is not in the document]
paouelle: However, the manager is the one taking this property and passing it to the services right now ... so it kinda care ;-)

query/src/main/java/com/connexta/replication/query/QueryService.java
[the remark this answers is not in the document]
paouelle: Oops! Nevermind

query/src/main/java/com/connexta/replication/query/QueryManager.java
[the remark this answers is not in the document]
paouelle: It might be safe to still handle them wherever you handle `AdapterException`. I would rather not rely on how the code of that other class is but more on the fact that it documents it can throw that base AdapterException which in the end, could be any subclasses.

query/src/main/java/com/connexta/replication/query/QueryServiceTools.java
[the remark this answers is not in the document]
kcover: hmm, I created QueryServiceTools to hold the services etc. that would need to be passed to both the services and the manager. Since the config polling period only pertains to the Manager I thought to leave it out. Either way this is a small change that I don't want to worry about right now.

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,421 @@
...
+        queue.waitForPendingSizeToReach(3, 5L, TimeUnit.SECONDS),
+        Matchers.equalTo(true));
+
+    for (Metadata metadata : metadataList) {
+      assertTask(queue.take(), metadata, (byte) 0, OperationType.HARVEST);
+    }
+    verify(indexManager)
+        .save(
+            Mockito.argThat(
+                isIndexWithModified(metadataList.get(2).getMetadataModified().toInstant())));
```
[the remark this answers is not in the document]
kcover: The metadata for this test are created in a list instead of as individual variables. This is just getting the last element in the list. 

I'll be sure to remove the spy. 🔭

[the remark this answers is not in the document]
paouelle: I knew where the `2` was coming from. :-)
 We might want to clarify that by defining a constant or something that by its name, clarify it is that particular entry in the list of metacards returned above.

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
kcover: This is the config polling period. Not the query polling period

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
paouelle: Can we create a ticket for this?

query/src/test/java/com/connexta/replication/query/cmp/QueryServiceComponentTest.java
```diff
@@ -0,0 +1,416 @@
...
+        Matchers.equalTo(true));
+    assertTask(regionalQueue.take(), metadataList.get(2), (byte) 0, OperationType.HARVEST);
+    assertThat(tacticalQueue.pendingSize(), Matchers.equalTo(0));
+    assertThat(ionQueue.pendingSize(), Matchers.equalTo(0));
+
+    assertIndexSaved(filterList.get(2), metadataList.get(2).getMetadataModified());
+  }
+
+  private void assertIndexSaved(Filter filter, Date lastModified) {
+    Awaitility.await().until(() -> indexManager.getOrCreate(filter).getModifiedSince().isPresent());
```
paouelle: ❗️ We might want to put a timeout here so that if it doesn't happen in 5-10 seconds, we fail the test case instead of letting it be stuck.

paouelle:
🎉 Nice job on this PR

commons/src/main/java/com/connexta/replication/data/MetadataImpl.java
```diff
@@ -51,17 +51,39 @@
...
+   * @param metadata the raw metadata to wrap, cannot be null
+   * @param type the type of the metadata, cannot be null
+   * @param id id the of the metadata, cannot be null or empty
+   * @param metadataModified the last time the metadata was modified, cannot be null
+   * @throws NullPointerException if any of the parameters are null
+   * @throws IllegalArgumentException if ID is empty or the type doesn't match the metadata
+   */
+  public MetadataImpl(
+      long metadataSize, Object metadata, Class type, String id, Date metadataModified) {
+    if (!type.isInstance(metadata)) {}
```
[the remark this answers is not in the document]
kcover: yea not sure how I forgot to add a body to that 🤔 . Thanks.

query/src/main/java/com/connexta/replication/query/QueryManager.java
```diff
@@ -0,0 +1,199 @@
...
+      }
+      services.clear();
+    }
+  }
+
+  @VisibleForTesting
+  void reloadSiteConfigs() {
+    Stream<Site> siteStream = queryServiceTools.allSites();
+    synchronized (this) {
+      if (Thread.currentThread().isInterrupted()) {
```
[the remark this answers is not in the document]
kcover: The manager will only be interrupted if it's being shutdown. The reason this block is synchronized and this check is here is because if a thread calls this method while another thread calls destroy, a new service can be created after they should all have been shutdown. This check makes sure that if we were shutdown we don't create anymore QueryServices.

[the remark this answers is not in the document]
paouelle: @peterhuffer he check done here doesn't affect the interruption state of the thread. Normally the executor would automatically let the thread die and create a new one. However, as Kiefer mentioned, interruptions don't magically happen and in this case, they would happen because of the shutdown done in the `destroy()` method which means the manager is shutting down.

query/src/main/java/com/connexta/replication/query/QueryServiceTools.java
```diff
@@ -0,0 +1,146 @@
...
+    return filterIndexManager.getOrCreate(filter);
+  }
+
+  /**
+   * Persists the given filter index.
+   *
+   * @param filterIndex the filter index to persist
+   */
+  public void saveFilterIndex(FilterIndex filterIndex) {
+    filterIndexManager.save(filterIndex);
```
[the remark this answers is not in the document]
paouelle: Good point. All these methods should document that these exceptions can be thrown and based on following the code, nothing is handling these exceptions which means it will abort everything. We need to add catches in various places where we are calling these persistence method to handle these exceptions (i.e. `ReplicationException`).

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -0,0 +1,278 @@
...
+      final FilterIndex filterIndex = queryServiceTools.getOrCreateFilterIndex(filter);
+
+      // holds the latest modified time received so far
+      AtomicReference<Instant> latestModifiedTime =
+          new AtomicReference<>(filterIndex.getModifiedSince().orElse(Instant.MIN));
+
+      try {
+        getNewMetadataForFilter(filter, filterIndex)
+            .sorted(Comparator.comparing(Metadata::getMetadataModified))
+            .map(metadatum -> metadataToTask(metadatum, filter.getPriority()))
```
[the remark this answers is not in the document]
kcover: Datum is the singular form of data. Metadatum isn't actually a word though so I think I'll change this. It's not communicating anything now like it was before anyway.

PR #230, opened 2019-08-29
Add worker manager and workers

peterhuffer:
#### What does this PR do?
Builds off of https://github.com/connexta/replication/pull/229

Only review https://github.com/connexta/replication/pull/230/commits/18a33a37b9ada4006681573354c977d1f646075b . Other commits are from PR 229.

Adds a manager which monitors sites and spins up/down workers based on the site's configuration

@paouelle @clockard @kcover 

CI Build.

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests


peterhuffer:
I need to make NodeAdapters into an interface until I can move my code into a separate module.

replication-api/src/main/java/com/connexta/replication/api/data/ErrorCode.java
```diff
@@ -21,7 +21,10 @@
   SITE_TIMEOUT(true),
   UPSTREAM_SERVICE_UNAVAILABLE(true),
   UPSTREAM_SERVICE_TIMEOUT(true),
-  UNKNOWN_ERROR(false);
+  UNKNOWN_ERROR(false),
+  UNSUPPORTED_ERROR(true),
+  // TODO: Using this for now until adapters can report better errors
```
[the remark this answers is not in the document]
peterhuffer: I think that this should be done in a separate ticket since it is a limitation of the adapters right now.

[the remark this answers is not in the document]
paouelle: I am good with that but see other comment to remove the UNSUPPORTED one.

replication-api/src/main/java/com/connexta/replication/api/data/ErrorCode.java
```diff
@@ -21,7 +21,10 @@
   SITE_TIMEOUT(true),
   UPSTREAM_SERVICE_UNAVAILABLE(true),
   UPSTREAM_SERVICE_TIMEOUT(true),
-  UNKNOWN_ERROR(false);
+  UNKNOWN_ERROR(false),
+  UNSUPPORTED_ERROR(true),
```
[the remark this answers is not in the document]
peterhuffer: I was using this for a worker that received an unknown operation type, but will switch it to unlock the task and get rid of this based on discussions earlier.
❓ How will the queue know not to send the same task back to the worker that just received it even though the worker could not handle it?

paouelle:
I think moving the worker into a separate module can be done once we work on micro services for them at which point they will actually move to a different repo.

replication-api/src/main/java/com/connexta/replication/api/data/ErrorCode.java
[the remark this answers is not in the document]
peterhuffer: If we are alright with that I can make the ticket and link it here.

replication-api/src/main/java/com/connexta/replication/api/data/ErrorCode.java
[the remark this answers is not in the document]
paouelle: Really good question and for now, it won't. For the monolithic service, that will never happen. For the micro services one, the deployment sequences should already have started newer workers and would most likely be in the process shutting down the older ones. So worst case scenario, it gets picked up again by the same until shut down.

Once we have integrated with the external queue, we can see if there are ways to mark it as such that it wouldn't be picked up again by the same worker but based on how we would get there and what I said above, I am not sure we should worry about it.

replication-api/src/main/java/com/connexta/replication/api/data/ErrorCode.java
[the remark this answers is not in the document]
paouelle: I am ok with your statement. That won't be true once we talk about interruptions later where I don't think we can push enhancing the adapter to properly support them based on the need of the worker and the query service.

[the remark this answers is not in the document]
peterhuffer: Made an issue here https://github.com/connexta/replication/issues/235

replication-api/src/main/java/com/connexta/replication/api/data/ddf/DDFMetadataInfo.java
```diff
@@ -25,12 +25,12 @@
    *
    * @return the class for the raw data
    */
-  public Class<?> getDataClass();
+  Class<?> getDataClass();
```
paouelle: 😄

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+        task.fail(ErrorCode.SITE_UNAVAILABLE,
+            String.format("Site %s is unavailable", local.getSystemName()));
+      }
+
+      if (!destination.isAvailable()) {
+        task.fail(ErrorCode.SITE_UNAVAILABLE,
+            String.format("Site %s is unavailable", destination.getSystemName()));
+      }
+    } catch (AdapterException e) {
+      LOGGER.debug("Error retrieving availability for adapters", e);
```
paouelle: ❓ Shouldn't we handle those as failures of some sort?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+    }
+    // if the task didn't complete with the call to fail, we still have the lock
+    return task.isLocked();
```
paouelle: ❓ Why not simply return false right away above when you fail the task and here return true? Your logic knows here one way or anther and you shouldn't need to check with the task.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+    if (!OperationType.UNKNOWN.equals(task.getOperation())) {
+      task.fail(ErrorCode.UNSUPPORTED_ERROR,
+          String.format("Operation task %s is not supported by this worker", task));
+      return false;
+    }
+    return true;
+  }
+
+  private void handle(Task task) throws InterruptedException {
+    boolean proceed = checkAdapterAvailabilities(task);
```
paouelle: ❓ Do we need to store the result? Why not just check in a `if` and return right away?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+    if (!proceed) {
+      return;
+    }
+
+    List<DDFMetadataInfo> infos = task.metadatas()
```
paouelle: ❗️ The list should be `List<MetadataInfo>` since you don't know yet what type it is

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+        .filter(t -> t instanceof DDFMetadataInfo)
```
paouelle:
```suggestion
        .filter(DDFMetadataInfo.class::isInstance)
```
❓ How will that work later when dealing with Ion? If we don't support it yet (which I know we don't) then we should make sure to filter these tasks out as soon as possible and let someone else handle them (like a newer version). This check shuld be added to your `supported()` method.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+        .map(DDFMetadataInfo.class::cast)
+        .collect(Collectors.toList());
+
+    if (infos.size() > 1) {
```
paouelle: ❗️ We shouldn't fail here but instead filter those out in the `supported()` method such that anewer worker can pick it up when we do add support for it.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+      task.fail(ErrorCode.UNSUPPORTED_ERROR,
+          "more than one metadata is currently not supported by this worker");
+      return;
+    }
+
+    if(infos.isEmpty()) {
```
paouelle: ❗️ Same comment about not failing but instead filtering those out.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+      task.fail(ErrorCode.UNSUPPORTED_ERROR,
+          "unable to handle any of the metadatas received in this task");
+      return;
+    }
+
+    final DDFMetadataInfo info = infos.get(0);
```
paouelle: ❓ Should this worker be more generic and not care about DDF metadata info but instead pass that info to the adapter and let the adapter handle it properly. A ddf adapter would care and on a query, it would simply return it back whereas as another one would perform its query to get the data.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+        break;
+      case REPLICATE_TO:
+        doCreate(task, metadata, local, destination);
+        break;
+      case DELETE:
+        // always delete on non-local site
+        doDelete(task, metadata);
+        break;
+      default:
+        // should never hit this since deserialization will return UNKNOWN
```
paouelle: ❓ Same comment as before, this should be handled in the `supported() ` method to let another worker (a newer one) pick up something you cannot do.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+        task.fail(ErrorCode.OPERATION_FAILURE,
+            String.format("Failed to %s metadata %s from %s to %s",
+                operation,
+                metadataId,
+                source.getSystemName(),
+                destination.getSystemName()));
+      }
+    } else {
+      LOGGER.info("Received task to create metadata on site {}, but it was deleted on {}",
+          destination.getSystemName(), destination.getSystemName());
```
paouelle: ❓ Should the second one be `source`?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+      task.complete();
+    }
+
+    ResourceInfo resourceInfo = task.getResource().orElse(null);
```
paouelle: ✏️ You could combine these 2 `if`s into one by doing:

`URI uri = task.getResource().flatMap(ResourceInfo::getResourceUri).orElse(null);`

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+                    destination.getSystemName()));
+          }
+        } catch (AdapterException e) {
+          LOGGER.debug("Failed to retrieve resource", e);
+          task.fail(ErrorCode.OPERATION_FAILURE,
+              String.format("Failed to read resource from %s", destination.getSystemName()));
+        }
+      } else {
+        LOGGER.warn("No resource URI specified for task {}", task);
+        return;
```
paouelle: ❓ What is going to happen to the task?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+            String.format("Operation task %s is not supported by this worker", task));
+    }
+  }
+
+  private void doCreate(final Task task, final Metadata metadata, final NodeAdapter source,
+      final NodeAdapter destination)
+      throws InterruptedException {
+    final String metadataId = task.getId();
+    final String operation = task.getOperation().name();
+    if (source.exists(metadata)) {
```
paouelle: ❓ Are we going to always check first? Does that now mean that for all normal cases we shall end up with 2 requests to the source, one to check and one to get the data? Can we not take the approach where we try to get it and it either tells us it doesn't have it (e.g. 404-not found) or we get the data?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+      CreateRequest request = new CreateRequestImpl(List.of(metadata));
+      boolean success = destination.createRequest(request);
+      if (!success) {
+        task.fail(ErrorCode.OPERATION_FAILURE,
+            String.format("Failed to %s metadata %s from %s to %s",
+                operation,
+                metadataId,
+                source.getSystemName(),
+                destination.getSystemName()));
+      }
```
paouelle: ❗️ Should return to avoid the `task.complete()` below which will fail

paouelle: ❓ What happens to the task in the `else`?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+        doDelete(task, metadata);
+        break;
+      default:
+        // should never hit this since deserialization will return UNKNOWN
+        task.fail(ErrorCode.UNSUPPORTED_ERROR,
+            String.format("Operation task %s is not supported by this worker", task));
+    }
+  }
+
+  private void doCreate(final Task task, final Metadata metadata, final NodeAdapter source,
```
paouelle: ❗️ Exception handling when calling `exist()`, `createRequest`, and all other methods on the node adapter needs to happen somewhere. We also need to handle properly interruption caused by the manager trying to stop the worker which could translate into `InterruptedIOException` which needs to be properly reported by the adapters. And that I don't think can be pushed off to a different story.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+    if (source.exists(metadata)) {
+      CreateRequest request = new CreateRequestImpl(List.of(metadata));
+      boolean success = destination.createRequest(request);
+      if (!success) {
+        task.fail(ErrorCode.OPERATION_FAILURE,
+            String.format("Failed to %s metadata %s from %s to %s",
+                operation,
+                metadataId,
+                source.getSystemName(),
+                destination.getSystemName()));
```
paouelle: ❗️ Looks like it might fall-through below where the task will end up being marked completed. If it failed then that is it and we should return

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+          Resource resource = response.getResource();
+          boolean success = destination
+              .createResource(new CreateStorageRequestImpl(List.of(resource)));
+          if (!success) {
+            task.fail(ErrorCode.OPERATION_FAILURE,
+                String.format("Failed to %s resource %s from %s to %s",
+                    operation,
+                    uri,
+                    source.getSystemName(),
+                    destination.getSystemName()));
```
paouelle: ❗️ This will fall-through to the `task.complete()` below and fail. You have to return

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+   * Creates a new worker which will monitor the provided queue and perform operations between the
+   * provided local and destination sites.
+   *
+   * Does not begin monitoring of the queue until {@link #run()} is called.
+   *
+   * @param queue the queue to pull tasks from; corresponding queue for the destination site
+   * @param local the local site adapter
+   * @param destination the destination site adapter
+   */
+  public Worker(Queue queue, NodeAdapter local, NodeAdapter destination) {
```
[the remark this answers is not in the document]
peterhuffer: I was trying to avoid passing down the NodeAdapters class but I guess I have to.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+              .createResource(new CreateStorageRequestImpl(List.of(resource)));
+          if (!success) {
+            task.fail(ErrorCode.OPERATION_FAILURE,
+                String.format("Failed to %s resource %s from %s to %s",
+                    operation,
+                    uri,
+                    source.getSystemName(),
+                    destination.getSystemName()));
+          }
+        } catch (AdapterException e) {
```
paouelle: ❗️ As stated before, We will need to handle interruptions that stops the adapter since that is not an error.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+  private void doDelete(Task task, Metadata metadata) throws InterruptedException {
+    final String metadataId = task.getId();
+    final String operation = task.getOperation().name();
+    boolean success = destination.deleteRequest(new DeleteRequestImpl(List.of(metadata)));
+    if (!success) {
+      task.fail(ErrorCode.OPERATION_FAILURE,
+          String.format("Failed to %s metadata %s on %s",
+              operation,
+              metadataId,
+              destination.getSystemName()));
```
paouelle: ❗️ Need to return to avoid the `task.complete()` below

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+        return;
+      }
+    }
+    task.complete();
+  }
+
+  private void doDelete(Task task, Metadata metadata) throws InterruptedException {
+    final String metadataId = task.getId();
+    final String operation = task.getOperation().name();
+    boolean success = destination.deleteRequest(new DeleteRequestImpl(List.of(metadata)));
```
paouelle: ❗️ Need to handle adapter exception.
❓ What happens if it doesn't exist on the site? Can we detect that type of failure from the adapter and throw a different error so that here we do not have to fail since it is a no-op and therefore the task should be completed.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
```diff
@@ -0,0 +1,123 @@
...
+      try {
+        siteManager.objects().forEach(this::monitorSite);
+      } catch (ReplicationPersistenceException e) {
+        LOGGER.debug("Error retrieving sites", e);
+        LOGGER.error("Error retrieving sites to monitor, will try again next poll interval.");
+      }
+    } else {
+      for (String siteId : siteIds) {
+        try {
+          Site site = siteManager.get(siteId);
```
paouelle: ✏️ Would be more optimal to get all sites and filter the ones you don't want.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
```diff
@@ -0,0 +1,123 @@
...
+
+  private final Set<String> siteIds;
+
+  private final Site localSite;
+
+  private final SiteManager siteManager;
+
+  private final NodeAdapters nodeAdapters;
+
+  private final Map<String, WorkerThreadPool> threadPools = new HashMap<>();
```
paouelle: ❗️ Should be a concurrent one.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }
+  }
+
+  private void monitorSite(Site site) {
+    final String siteId = site.getId();
+
+    final int poolSize = Math
+        .min(localSite.getParallelismFactor().orElse(1), site.getParallelismFactor().orElse(1));
+    WorkerThreadPool threadPool = threadPools.get(siteId);
```
paouelle: ✏️ This can be achieve with:
```
threadPools.compute(siteId,this::createOrUpdate);

private WorkerThreadPool createOrUpdate(String site, @Nullable WorkerThreadPool current) {
   if (current == null) {
      return new WorkerThreadPool(broker.getQueue(siteId), poolSize, localSite, site,
          nodeAdapters);
   }
   current.setSize(poolSize);
   return current;
}
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerThreadPool.java
```diff
@@ -0,0 +1,160 @@
...
+
+    final int delta = currentSize - size;
+    spinDownWorkers(delta);
+  }
+
+  private void spinDownWorkers(int target) {
+    int cancelled = 0;
+    // try to cancel non processing workers first
+    for (Worker worker : workers) {
+      if (!worker.isProcessing()) {
```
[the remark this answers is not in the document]
peterhuffer: Do I need a lock if the `processing` var is volatile?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
```diff
@@ -0,0 +1,123 @@
...
+        }
+      }
+    }
+  }
+
+  private void monitorSite(Site site) {
+    final String siteId = site.getId();
+
+    final int poolSize = Math
+        .min(localSite.getParallelismFactor().orElse(1), site.getParallelismFactor().orElse(1));
```
paouelle: ❗️ The parallelism factor should not be based on the site and not the local one. We need a default to be configurable so we fallback to a default but that should not be the one for the local site.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerThreadPool.java
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * Manages a site's set of workers which in turn run workers.
+ */
+public class WorkerThreadPool {
+
+  private final Object lock = new Object();
+
+  private boolean shutdown = false;
+
+  private final List<Worker> workers = Collections.synchronizedList(new ArrayList<>());
```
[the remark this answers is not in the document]
peterhuffer: Switched to CopyOnWriteArrayList

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerThreadPool.java
```diff
@@ -0,0 +1,160 @@
...
+   */
+  public WorkerThreadPool(Queue queue, int size,
+      Site localSite,
+      Site remoteSite,
+      NodeAdapters nodeAdapters) {
+    this.queue = queue;
+    this.localSite = localSite;
+    this.destinationSite = remoteSite;
+    this.nodeAdapters = nodeAdapters;
+    setSize(size);
```
paouelle: ❗️ A public method should never be called from within the constructor unless it is final

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Class responsible for managing item workers. It will monitor sites currently available in storage
+ * and create/destroy item workers accordingly for each site, respecting each site's parallelism
+ * factor.
+ */
+public class WorkerManager {
```
paouelle: ❓ Should this bean have a destroy() method such that spring can shut it down? This is where you were turn around and shutdown all pools.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerThreadPool.java
```diff
@@ -0,0 +1,160 @@
...
+   * @throws IllegalStateException if this thread pool has been shutdown
+   */
+  public void setSize(int size) {
+    synchronized (lock) {
+      if (shutdown) {
+        throw new IllegalStateException("This thread is shutdown");
+      }
+    }
+
+    final int currentSize = this.getSize();
```
paouelle: ❗️ The whole operations of checking the size and updating it should be done under lock to prevent other threads from possibly calling this method while you are affecting the state of the object stored in multiple attributes.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerThreadPool.java
```diff
@@ -0,0 +1,160 @@
...
+   * If the new size is greater than the current size, then newSize-currentSize workers will be spun
+   * up.
+   *
+   * @param size the size of this thread pool
+   * @throws IllegalStateException if this thread pool has been shutdown
+   */
+  public void setSize(int size) {
+    synchronized (lock) {
+      if (shutdown) {
+        throw new IllegalStateException("This thread is shutdown");
```
paouelle: ✏️ It might actually happen that you have a race condition between a shutdown to the manager's bean and the pool actually being shutdown. Should we simply return as if nothing happened?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerThreadPool.java
```diff
@@ -0,0 +1,160 @@
...
+        if (cancelled == target) {
+          break;
+        }
+      }
+    }
+  }
+
+  private void spinUpWorkers(int num) {
+    for (int i = 0; i < num; i++) {
+      NodeAdapter localAdapter = nodeAdapters.factoryFor(localSite.getType())
```
paouelle: ✏️ Since when we move to external microservices, the worker will be responsible for getting their own adapter(s). I would suggest we do not do this here and leave it up to the constructor of the Worker to do that.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  /**
+   * Initializes this manager and begins monitoring of sites.
+   */
+  public void init() {
+    executorService.scheduleAtFixedRate(this::monitorSites, 0L, 30L, TimeUnit.SECONDS);
+  }
+
+  private void monitorSites() {
```
paouelle: ❗️ We should also make sure to filter any sites we don't support (not just those not configured in `sites`). The kind and the type of the site might be unknown if this is an older version of the manager.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerThreadPool.java
```diff
@@ -0,0 +1,160 @@
...
+
+  /**
+   * Shutdown this thread pool, interrupting every worker and forbidding new workers from being
+   * added.
+   */
+  public void shutdown() {
+    synchronized (lock) {
+      shutdown = true;
+    }
+    workers.forEach(Thread::interrupt);
```
paouelle: ❓ Should we not remove them from the list?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,259 @@
...
+   *
+   * @param queue the queue to pull tasks from; corresponding queue for the destination site
+   * @param local the local site adapter
+   * @param destination the destination site adapter
+   */
+  public Worker(Queue queue, NodeAdapter local, NodeAdapter destination) {
+    this.queue = queue;
+    this.local = local;
+    this.destination = destination;
+  }
```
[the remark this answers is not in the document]
peterhuffer: ❓ It sounds like if the main user thread shuts down (our worker manager thread in this case) that the daemon threads will terminate automatically, will that cause problems for unlocking the task in the queue?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: 👍 Put it as an unknown error.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: I know they are since it filters with an instanceof.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Added to supported check and removed failure and replaced with unlock

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Added to supported check and removed failure and replaced with unlock

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: I switched this to unlock instead of fail but going to keep this default case here just in case and sonarlint likes it there.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Whoops, I'll let it fall through to the complete call below.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: It would fall down to the end of the method and complete, but I do need to return here to prevent calling fail then complete.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Caught adapter exception and for now did an exists check and left a todo to remove once adapters support better error reporting

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
peterhuffer: Switched this to query the sites and get the difference in the set of sites i have in my thread pool map, then shutdown thread pools that no longer have a corresponding site in storage

paouelle: Don't forget to filter on the set of site ids configured, those of unknown type and kind, ...

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerThreadPool.java
peterhuffer: Made it final for now, but i don't see this class ever being extended

paouelle: The point is that you should never call a method that can be overridden (wether you think it will or will never be) since you are calling into code of a class that hasn't been fully constructed yet. Even doing that somewhere in the constructor before all attributes have been initialized can create weird things since your method might be expecting an attribute to be null and it might be if it wasn't initialized yet. Even worst for derived classes since those have not be constructed yet and would get their overridden method called.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Kept the exist and left a todo with a link to https://github.com/connexta/replication/issues/235

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Yeah I think eventually we will want that, but I'd like to avoid modifications to the adapter API in this PR if we can defer it.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Added the unknown, size and type checks to the supported method and replaced the failures with unlocks.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,335 @@
...
+      return;
+    }
+
+    List<DDFMetadataInfo> infos = task.metadatas()
+        .filter(DDFMetadataInfo.class::isInstance)
+        .map(DDFMetadataInfo.class::cast)
+        .collect(Collectors.toList());
+
+    final DDFMetadataInfo info = infos.get(0);
+    final Metadata metadata = new MetadataImpl(((Map) info.getData()).get("raw-metadata"),
```
clockard: This will change when https://github.com/connexta/replication/pull/215 goes in. It should simplify this I think.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,335 @@
...
+      final NodeAdapter destination)
+      throws InterruptedException {
+    final String metadataId = task.getId();
+    final String operation = task.getOperation().name();
+
+    try {
+      // TODO: remove existence check once better reporting is done by adapters
+      // see https://github.com/connexta/replication/issues/235
+      if (source.exists(metadata)) {
+        CreateRequest request = new CreateRequestImpl(List.of(metadata));
```
clockard: ❗️ Doing separate metadata and resource creates will not work with any DDF based system. It doesn't really work with Ion either but it the adapter just quietly fails for metadata creates. Ion only supports resource+metadata and DDF only supports metadata only, or both resource + metadata in one request.

clockard: I know this concept was talked about a little bit but we are not ready for that yet. We can always add this functionality later if needed and the systems can support it.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,335 @@
...
+    } catch (AdapterException ae) {
+      LOGGER.debug("Failed to retrieve resource", ae);
+      task.fail(ErrorCode.OPERATION_FAILURE,
+          String.format("Failed to read resource from %s", destination.getSystemName()));
+      return;
+    }
+
+    URI uri = task.getResource().flatMap(ResourceInfo::getResourceUri).orElse(null);
+    if (uri != null) {
+      ResourceRequest request = new ResourceRequestImpl(metadata);
```
clockard: This is the only request we should support right now when dealing with Ion and existing DDFs.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,335 @@
...
+    final DDFMetadataInfo info = infos.get(0);
+    final Metadata metadata = new MetadataImpl(((Map) info.getData()).get("raw-metadata"),
+        info.getDataClass(), task.getId(),
+        Date.from(info.getLastModified()));
+
+    final OperationType type = task.getOperation();
+    switch (type) {
+      case HARVEST:
+      case REPLICATE_FROM:
+        doCreate(task, metadata, destination, local);
```
clockard: If the query service still works how I think it is then we will also get updates via the queue as well. I assume we are expecting the create to fail if we have already created it (based on correlation id). How are we going to handle updates that come through and just keep failing? Might not be a concern for the worker but we need to make sure we handle that. Ion needs to probably handle updates to solve this.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Okay, I can change this to check if a resource exists first, and if it does do the resource create, otherwise do a metadata create.

peterhuffer: Or aren't we always only harvesting from DDF based systems? @paouelle

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: AFAIK updates will not be supported in this iteration. The query service does not have an appropriate operation type to use.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
paouelle: at the moment, we are only harvesting from regional DDF systems as far as I know. This should be ensured by the query service.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
paouelle: updates should be treated down the road as upserts. This is purposely why we don't differentiate updates from creates in the type of operations. For DDF, it would be up to the worker to handle that by updating first and if it fails creating or vice versa. We cannot really know in advance since the data can always arrive at a destination via other sites in the ecosystem.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Actually based on our discussion yesterday, if we want to make DDF the local site in order to continue support for the E2E tests, I will have to make this change.

peterhuffer: Updated.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,377 @@
...
+      site = siteManager.get(siteId);
+    } catch (NotFoundException e) {
+      LOGGER.debug("Unable to find site {} for queue", siteId, e);
+      return null;
+    } catch (ReplicationPersistenceException e) {
+      LOGGER.debug("Failed to fetch site {} for queue", siteId, e);
+      return null;
+    }
+
+    return nodeAdapters.factoryFor(site.getType()).create(site.getUrl());
```
paouelle: ❗️ `getAdapterFor()` can throw `IllegalArgumentException` if it is unable to find a factory for the provided site.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,377 @@
...
+    if (ddfInfo == null) {
+      LOGGER.info("Unable to handle any of the metadatas received in this task");
+      return false;
+    }
+
+    return true;
+  }
+
+  private void handle(Task task) throws InterruptedException {
+    NodeAdapter destination = getDestinationAdapter(task);
```
paouelle: ❗️ `getDestinationAdapter()` can throw `IllegalArgumentException` if it is unable to find a factory for the provided site.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: Caught it and AdapterException and unlocked task if either happens.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
```diff
@@ -0,0 +1,148 @@
...
+  @VisibleForTesting
+  void monitorSites() {
+    Stream<Site> supportedSites =
+        siteManager
+            .objects()
+            .filter(site -> !SiteType.UNKNOWN.equals(site.getType()))
+            .filter(site -> !SiteKind.UNKNOWN.equals(site.getKind()))
+            .filter(site -> !site.getId().equals(localSite.getId()));
+
+    if (siteIds.isEmpty()) {
```
paouelle:
```
    if (!siteIds.isEmpty()) {
      supportedSites = supportedSites.filter(site -> siteIds.contains(site.getId()));
    }
    supportedSites.forEach(this::monitorSite);
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
```diff
@@ -0,0 +1,148 @@
...
+      supportedSites.forEach(this::monitorSite);
+    } else {
+      supportedSites.filter(site -> siteIds.contains(site.getId())).forEach(this::monitorSite);
+    }
+
+    // if a site has been removed from storage, cleanup its thread pool
+    Set<String> threadPoolSites = new HashSet<>(threadPools.keySet());
+    siteManager.objects().forEach(site -> threadPoolSites.remove(site.getId()));
```
paouelle: ❗️ We shouldn't be going back to Solr twice. The logic here should be based on one snapshot of sites retrieved from Solr.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/WorkerManager.java
```diff
@@ -0,0 +1,148 @@
...
+    if (threadPool != null) {
+      threadPool.setSize(poolSize);
+    } else {
+      threadPool = threadPoolFactory.create(siteId, poolSize);
+      threadPools.put(siteId, threadPool);
+    }
+  }
+
+  private void shutdownThreadPool(String siteId) {
+    WorkerThreadPool threadPool = threadPools.get(siteId);
```
paouelle:
```suggestion
    WorkerThreadPool threadPool = threadPools.remove(siteId);
```

Then you don't need to remove it after. You end up accessing the collection once instead of twice.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,405 @@
...
+      }
+      this.processing = false;
+    } catch (InterruptedException e) {
+      if (task != null) {
+        try {
+          task.unlock();
+        } catch (InterruptedException ignore) { // fallthrough
+        }
+      }
+      running = false;
```
paouelle: ❓ Should this line actually be put after the while loop in `run()`?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,405 @@
...
+  boolean cancelIfNotProcessing() {
+    if (!processing) {
+      this.running = false;
+      return true;
+    }
+    return false;
+  }
+
+  private boolean supported(Task task) {
+    if (OperationType.UNKNOWN.equals(task.getOperation())) {
```
paouelle:
```suggestion
    if (OperationType.UNKNOWN == task.getOperation()) {
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,405 @@
...
+      return false;
+    }
+
+    List<MetadataInfo> infos = task.metadatas().collect(Collectors.toList());
+    if (infos.size() > 1) {
```
paouelle:
```suggestion
   return (task.metadatas().filter(DdfMetadataInfo.class::isInstance).count() > 1) {
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
paouelle: ❓ Should we actually filter out all non-harvest tasks for now?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,405 @@
...
+      LOGGER.warn("More than one metadata is currently not supported by this worker");
+      return false;
+    }
+
+    DdfMetadataInfo ddfInfo =
+        infos.stream()
+            .filter(DdfMetadataInfo.class::isInstance)
+            .map(DdfMetadataInfo.class::cast)
```
paouelle: ✏️ If you end up keeping this part of the method, there is no need to case and no need to find first. Just do a `count()`

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,405 @@
...
+  }
+
+  @Nullable
+  private NodeAdapter getDestinationAdapter(Task task) {
+    Site site;
+    final String siteId = task.getQueue().getSite();
+    try {
+      site = siteManager.get(siteId);
+    } catch (NotFoundException e) {
+      LOGGER.debug("Unable to find site {} for queue", siteId, e);
```
paouelle: ❓ Should we be failing the task in such a way that it won't be retried if the site no longer exists? Or again completed as a no-op?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,405 @@
...
+    try {
+      if (!adaptersAvailable(task, remote)) {
+        return;
+      }
+
+      List<DdfMetadataInfo> infos =
+          task.metadatas()
+              .filter(DdfMetadataInfo.class::isInstance)
+              .map(DdfMetadataInfo.class::cast)
+              .collect(Collectors.toList());
```
paouelle:
```suggestion
              .findFirst().get(); // should never end up being empty since it was filtered above
```

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,405 @@
...
+            "Received task to create metadata on site {}, but it was deleted on {}",
+            destination.getSystemName(),
+            source.getSystemName());
+        task.fail(
+            ErrorCode.NO_LONGER_EXISTS,
+            String.format(
+                "Metadata deleted from site %s before able to create", source.getSystemName()));
+        return;
+      }
+    } catch (AdapterException e) {
```
paouelle: ❗️ Need to handle `AdapterInterruptedExeption` on all methods from adapters

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,405 @@
...
+    return true;
+  }
+
+  @VisibleForTesting
+  void setProcessing(boolean processing) {
+    this.processing = processing;
+  }
+
+  @VisibleForTesting
+  boolean isRunning() {
```
paouelle: ✏️ You could make these method public.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerTest.java
```diff
@@ -0,0 +1,1186 @@
...
+    // setup
+    Queue queue = mock(Queue.class);
+
+    Worker worker =
+        new Worker(
+            queue, mock(NodeAdapter.class), mock(SiteManager.class), mock(NodeAdapters.class));
+    worker.setProcessing(true);
+
+    // when
+    worker.cancelIfNotProcessing();
```
paouelle: ❓ Should you validate the returned result?

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerTest.java
```diff
@@ -0,0 +1,1186 @@
...
+import java.util.stream.Stream;
+import org.codice.junit.rules.ClearInterruptions;
+import org.junit.Rule;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.MethodSource;
+
+public class WorkerTest {
```
paouelle: ❗️ I would want to see tests cases that shows that if the worker is interrupted, it stops its current processing properly. Interruption is part of the contract for a worker so we need to make sure that for one, the task currently being handled is unlocked, completed, or failed and that the interrupted state of the thread is propagated once the loop finishes.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,354 @@
...
+            destination.getSystemName(),
+            source.getSystemName());
+        task.fail(
+            ErrorCode.NO_LONGER_EXISTS,
+            String.format(
+                "Metadata deleted from site %s before able to create", source.getSystemName()));
+        return;
+      }
+    } catch (AdapterInterruptedException e) {
+      throw new InterruptedException(e.getMessage());
```
paouelle: ✏️ Don't bother with the message since it won't ever have one. Either create it blank or pass the exception as a cause only.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    URI uri = task.getResource().flatMap(ResourceInfo::getResourceUri).orElse(null);
+    if (uri != null) {
+      ResourceRequest request = new ResourceRequestImpl(metadata);
+      ResourceResponse response;
+      try {
+        response = source.readResource(request);
+      } catch (AdapterInterruptedException e) {
+        throw new InterruptedException(e.getMessage());
```
paouelle: ✏️ Don't bother with the message since it won't ever have one. Either create it blank or pass the exception as a cause only.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,354 @@
...
+        if (!success) {
+          task.fail(
+              ErrorCode.OPERATION_FAILURE,
+              String.format(
+                  "Failed to %s resource %s from %s to %s",
+                  operation, uri, source.getSystemName(), destination.getSystemName()));
+          return;
+        }
+      } catch (AdapterInterruptedException e) {
+        throw new InterruptedException(e.getMessage());
```
paouelle: ✏️ Don't bother with the message since it won't ever have one. Either create it blank or pass the exception as a cause only.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,354 @@
...
+        if (!success) {
+          task.fail(
+              ErrorCode.OPERATION_FAILURE,
+              String.format(
+                  "Failed to %s metadata %s from %s to %s",
+                  operation, metadataId, source.getSystemName(), destination.getSystemName()));
+          return;
+        }
+      } catch (AdapterInterruptedException e) {
+        throw new InterruptedException(e.getMessage());
```
paouelle: ✏️ Don't bother with the message since it won't ever have one. Either create it blank or pass the exception as a cause only.

replication-api-impl/pom.xml
```diff
@@ -118,6 +118,22 @@
...
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-starter-test</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.solr</groupId>
```
paouelle: ❓ Is this dependency only require for testing? Should it be marked with test scope?

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,354 @@
...
+      if (supported(task)) {
+        this.handle(task);
+      } else {
+        task.unlock();
+      }
+      this.processing = false;
+    } catch (InterruptedException e) {
+      if (task != null) {
+        try {
+          task.unlock();
```
paouelle: ✏️ To avoid the nested try/catch, it might be worthwhile to add a new `unlockUnterruptibly()` method to the task which would do exactly that or again renamed the existing `unlock()` to `unlockInterruptibly()` and add a new `unlock()` that would do exactly that

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
paouelle: I don't think you want the broker to have knowledge of the queues it maintains and what to do in such case with the remaining tasks. The broker should just be there to provide access to the queues and let whoever puts stuff in or pulls stuff from them decide. Having to many cooks in the kitchen just complicates everything.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,354 @@
...
+      return null;
+    } catch (ReplicationPersistenceException e) {
+      LOGGER.debug("Failed to fetch site {} for queue", siteId, e);
+      return null;
+    }
+
+    NodeAdapter nodeAdapter;
+    try {
+      nodeAdapter = nodeAdapters.factoryFor(site.getType()).create(site.getUrl());
+    } catch (IllegalArgumentException | AdapterException e) {
```
paouelle: ❗️ Handle `AdapterInterruptedExcetion`

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
```diff
@@ -0,0 +1,354 @@
...
+  private void doCreate(
+      final Task task,
+      final Metadata metadata,
+      final NodeAdapter source,
+      final NodeAdapter destination)
+      throws InterruptedException {
+    final String metadataId = task.getId();
+    final String operation = task.getOperation().name();
+
+    try {
```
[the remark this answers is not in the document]
paouelle: I am not sure it can be since the `AdapterInterruptedException` would be caught by the `catch(AdapterException)`. The only way I can think of is to create a functional interface that declares it throws `InterruptedException` and return some generic type. Then call whatever method throws an `AdapterException` through a new method that receives that functional interface that actually does the call and unwraps the `AdapterInterruptedException`. Something like this:

```
@FunctionalInterface
public interface InterruptableSupplier<T> {
    /**
     * Gets a result.
     *
     * @return a result
     */
    T get() throws InterruptedException;
}

private <T> T doInterruptibly(InterruptibleSupplier<T> s) {
  try {
    return s.get();
  } catch (AdapterInterrutedException e) {
    throw new InterruptedException(e);
  }
}
```
Then wrap one of the call below like this:
```
try {
  if (!doInterruptibly(() -> source.exists(metadata)) {
   ...
} catch (AdapterException e) {
...
}
```

You could even extend the `doInterruptily()` method to receive additional info or a `Runnable` to be called if an `AdapterException` is thrown before throwing back, thus removing the need to put a try/catch altogether here.

replication-api-impl/pom.xml
peterhuffer: It has a scope of test in the dependencyManagement section, does it need it here too?

paouelle: No. I looked but didn't see it. My mistake.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
peterhuffer: The interface does not declare it, should it?

pom.xml
```diff
@@ -447,6 +448,30 @@
...
+            <dependency>
+                <groupId>org.junit.jupiter</groupId>
+                <artifactId>junit-jupiter-engine</artifactId>
+                <version>${junit-jupiter.version}</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.junit.vintage</groupId>
+                <artifactId>junit-vintage-engine</artifactId>
+                <version>${junit-jupiter.version}</version>
```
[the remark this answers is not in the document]
peterhuffer: I didn't really want to affect the rest of the project. I figured fully transitioning over to Junit5 could be a separate issue.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerComponentTest.java
```diff
@@ -0,0 +1,170 @@
...
+    queue.put(taskInfo1);
+    queue.put(taskInfo2);
+    queue.put(taskInfo3);
+
+    assertTimeoutPreemptively(
+        Duration.ofSeconds(10),
+        () -> {
+          while (true) {
+            if (queue.size() == 3) {
+              break;
```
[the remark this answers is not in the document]
peterhuffer: This one is waiting for the total size. I think he was going to add an additional method for size.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/worker/Worker.java
paouelle: probably better to do it since you are designing this to handle interruptions. On top of it, it should be the same way to handle it as the rest.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerComponentTest.java
```diff
@@ -0,0 +1,160 @@
...
+    assertTimeoutPreemptively(
+        Duration.ofSeconds(10),
+        () -> {
+          while (true) {
+            if (queue.size() == 0) {
+              break;
+            }
+          }
+        });
+  }
```
paouelle: ❗️ Missing validation of outcome from your component. Like the fact that the tasks should be completed, failed, or re-queued. Like the fact that data in the tasks should be sent out to the destination site.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerComponentTest.java
```diff
@@ -0,0 +1,160 @@
...
+    TaskInfo taskInfo1 = createTaskInfo("taskId1");
+    TaskInfo taskInfo2 = createTaskInfo("taskId2");
+    TaskInfo taskInfo3 = createTaskInfo("taskId3");
+    queue.put(taskInfo1);
+    queue.put(taskInfo2);
+    queue.put(taskInfo3);
+
+    workerManager.init();
+
+    assertTimeoutPreemptively(
```
paouelle: ✏️ Can be replaced with the new `queue.waitForSizeToReach()`

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerComponentTest.java
```diff
@@ -118,43 +151,139 @@ public void cleanup() throws Exception {
...
-  public void testWorkerTakesFromQueue() throws Exception {
+  public void testWorkerHarvestDdfResource() throws Exception {
+    // setup
+    final URI resourceUri = new URI("http://remote.test:1234/testId");
+    final Date modifiedDate = new Date();
+    final int resourceSize = 10;
+    final Metadata metadata = createMetadata(METADATA_ID, resourceUri, modifiedDate, resourceSize);
+    final ResourceInfo resourceInfo = createResourceInfo(metadata);
+    final MetadataInfo metadataInfo = createMetadataInfo(metadata);
     MemorySiteQueue queue = (MemorySiteQueue) broker.getQueue(REMOTE_SITE_ID);
```
paouelle: ✏️ There is a new `getSiteQueue()` method that was made public

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerComponentTest.java
```diff
@@ -102,10 +135,10 @@ public void setup() {
     ReplicationProperties props = new ReplicationProperties();
     props.setLocalSite(LOCAL_SITE_ID);
 
-    final Site localSite = new SiteImpl(createSitePojo(LOCAL_SITE_ID, 1));
+    final Site localSite = new SiteImpl(createSitePojo(LOCAL_SITE_ID, LOCAL_SITE_URL, 1));
```
paouelle: ❓ Should we be testing with an Ion as our local site?

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerComponentTest.java
```diff
@@ -118,43 +151,139 @@ public void cleanup() throws Exception {
...
-    TaskInfo taskInfo1 = createTaskInfo("taskId1");
-    TaskInfo taskInfo2 = createTaskInfo("taskId2");
-    TaskInfo taskInfo3 = createTaskInfo("taskId3");
-    queue.put(taskInfo1);
-    queue.put(taskInfo2);
-    queue.put(taskInfo3);
+    TaskInfo taskInfo = createTaskInfo(METADATA_ID, metadataInfo, resourceInfo);
+    Task task = queue.putAndPeek(taskInfo);
 
+    ArgumentCaptor<CreateStorageRequest> createStorageRequestCaptor = ArgumentCaptor
```
paouelle: I can't wait to have our endpoint stubs so we can replace all that mocking with simple checks of what they received.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerComponentTest.java
```diff
@@ -118,43 +151,139 @@ public void cleanup() throws Exception {
...
-          while (true) {
-            if (queue.size() == 0) {
-              break;
-            }
-          }
-        });
+    // then
+    queue.waitForSizeToReach(0, 10, TimeUnit.SECONDS);
+
+    assertTrue(task.isCompleted());
```
paouelle: ❓ Should we be more specific by using `getState()` or `wasSuccessful()` or `hasFailed()` to check?

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerComponentTest.java
```diff
@@ -118,43 +151,139 @@ public void cleanup() throws Exception {
...
+    CreateRequest createRequest = createRequestCaptor.getValue();
+    Metadata sentMetadata = createRequest.getMetadata().get(0);
+    assertEquals(metadata.getType(), sentMetadata.getType());
+    assertEquals(metadata.getMetadataModified(), sentMetadata.getMetadataModified());
+    assertEquals(metadata.getRawMetadata(), sentMetadata.getRawMetadata());
+    assertEquals(metadata.getId(), sentMetadata.getId());
+  }
+
+  @Test
+  public void testMultipleTasksCompleted() throws Exception {
```
paouelle: ✏️ It be nice to test with one task that you know you will ignore and unlock and one task that fails

peterhuffer: I did this test mostly to check that a worker will continue to take from the queue after processing 1. I can add separate tests for an ignored and failed task though.

paouelle: I didn't mean in this test. For sure additional tests that verifies that would be great.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerThreadPoolTest.java
```diff
@@ -0,0 +1,152 @@
...
+import com.connexta.replication.api.persistence.SiteManager;
+import com.connexta.replication.api.queue.Queue;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+class WorkerThreadPoolTest {
```
paouelle: ✏️ `public`

peterhuffer: When I upgraded this module to use Junit5, it was no longer required so I just ended up removing it

PR #234, opened 2019-09-04
Java Doc Fix

clockard:
#### What does this PR do?
Fixes javadoc
Fixes an owasp finding
Updates some dependencies
@peterhuffer @paouelle @kcover 
CI Build


replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/AbstractPersistable.java
```diff
@@ -196,7 +196,7 @@ protected void setOrFailIfNullOrEmpty(
    *     destination
    * @param consumer a consumer capable of updating the destination with the field's value if it is
    *     defined
-   * @param <> the consumed type
+   * @param <P> the function and consumer type
```
paouelle: ✏️ It is not the function. It is only the consumed type by the function. The previous comment was correct except for the fact that it was missing the `T`

clockard: Fixed

PR #236, opened 2019-09-05
REP-225 Added support for interruptions to adapters 

paouelle:
#### What does this PR do?
This PR adds support for handling interruptions inside the adapters (DDF and Ion). It also removes the ability to create/update multiple resources which was not used and will not be used in the new architecture.

@peterhuffer 
@kcover 
@clockard 

CI

Fixes #XXXX

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/commons/src/main/java/com/connexta/replication/data/QueryResponseImpl.java
+++ b/commons/src/main/java/com/connexta/replication/data/QueryResponseImpl.java
@@ -15,18 +15,73 @@ package com.connexta.replication.data;
 
 import com.connexta.replication.api.data.Metadata;
 import com.connexta.replication.api.data.QueryResponse;
+import java.util.Iterator;
+import java.util.function.Function;
 
 /** Simple implementation of {@link QueryResponse}. */
 public class QueryResponseImpl implements QueryResponse {
 
   private final Iterable<Metadata> metadata;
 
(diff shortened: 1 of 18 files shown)


commons/src/main/java/com/connexta/replication/data/QueryResponseImpl.java
```diff
@@ -15,18 +15,73 @@
...
+    this.exceptionHandler = exceptionHandler;
+  }
+
+  private Iterator<Metadata> iterator() {
+    return new Iterator<>() {
+      final Iterator<Metadata> iterator = metadata.iterator();
+
+      @Override
+      public Metadata next() {
+        if (Thread.interrupted()) {
```
peterhuffer: ❓ If a caller calls hasNext before being interrupted, should we let them have the next element to let them finish what they were doing?

paouelle: What does this give us more? 

I only see that it might take a lot longer for them before they detect the interruption and handle it (assuming the processing of that item takes a long time). Typically, if you want to finish off, then you get interrupted, remember you were and go back to processing which wouldn't throw the interruption (unless somebody interrupts the thread again). You choose what you want to do once you are interrupted.

PR #257, opened 2019-09-25
REP-255 Wire together service architecture

### LINKED ISSUE #255: Get new service architecture implementations working together
**Describe the task**
Now that the functionality of replication has been separated into scalable services. Get the service modules working together to perform replication.

kcover:
#### What does this PR do?
Starts and connects the Query, Queue, and Worker services and
creates spring profiles for running the previous (Classic) and new (Ion)
architectures.

CI

Fixes #255 

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests


adapters/ddf-adapter/src/test/java/com/connexta/replication/adapters/ddf/DdfNodeAdapterTest.java
```diff
@@ -316,7 +316,110 @@ private Resource getResource() {
     Response response = mock(Response.class);
     when(status.getFamily()).thenReturn(failure ? Family.SERVER_ERROR : Family.SUCCESSFUL);
     when(response.getStatusInfo()).thenReturn(status);
-    when(function.apply(resource)).thenReturn(response);
+    when(function.apply(any())).thenReturn(response);
     return resource;
   }
+
+  private static final String RAW_METADATA =
```
kcover: Since we're using a static method to marshal/unmarshal metadata in the adapter we need to use real metadata to avoid errors.

commons/src/main/java/com/connexta/replication/data/ResourceImpl.java
```diff
@@ -97,4 +97,28 @@ public long getSize() {
   public Metadata getMetadata() {
     return metadata;
   }
+
+  @Override
+  public String toString() {
```
kcover: Added toString methods to this and MetadataImpl to help with logging.

distributions/test/src/test/java/com/connexta/replication/distributions/test/ClassicReplicationE2E.java
```diff
@@ -53,42 +55,56 @@
...
     assertTrue("Source and destination URLs not set!", sourceUrl != null && destinationUrl != null);
     LOGGER.info("Source URL: {}", sourceUrl);
     LOGGER.info("Destination URL: {}", destinationUrl);
     RestAssured.enableLoggingOfRequestAndResponseIfValidationFails();
-    replication.followOutput(logConsumer);
     solrUrl = "http://" + solr.getContainerIpAddress() + ":" + solr.getFirstMappedPort();
     LOGGER.info("Solr URL: {}", solrUrl);
+    String sites =
+        getFileAsText("sites.json", Map.of("source", sourceUrl, "destination", destinationUrl));
+    postToSolrCore(solrUrl, SITE, sites);
```
kcover: Replication will now break if it can't find the local site in solr so we have to make sure it's created before we start the replication container.

distributions/test/src/test/java/com/connexta/replication/distributions/test/IonReplicationE2E.java
```diff
@@ -0,0 +1,175 @@
...
+  private static final Slf4jLogConsumer logConsumer = new Slf4jLogConsumer(LOGGER);
+
+  @ClassRule public static Network network = Network.newNetwork();
+
+  private static final String sourceUrl = System.getenv("REPLICATION_SRC");
+  private static final String destinationUrl = System.getenv("REPLICATION_DEST");
+
+  public static GenericContainer solr = defaultSolrContainer(network);
+
+  public static GenericContainer replication = defaultReplicationContainer(network, "Ion");
```
kcover: This test class is the same as the Classic Tests with the exception of using the Ion profile and there's no deleted test case.

distributions/test/src/test/java/com/connexta/replication/distributions/test/TestUtils.java
```diff
@@ -86,7 +86,7 @@ public static GenericContainer defaultSolrContainer(Network network) {
         .withNetwork(network)
         .withNetworkAliases("replication-solr")
         .withExposedPorts(SOLR_PORT)
-        .waitingFor(Wait.forHttp("/solr/replication_" + FILTER + "/admin/ping"));
+        .waitingFor(Wait.forHttp("/solr/replication_" + SITE + "/admin/ping"));
```
kcover: Wanted to wait for the site core instead since it's the first core we call to.

distributions/test/src/test/resources/application.yml
```diff
@@ -20,4 +21,4 @@ replication:
   period: 1
   connectionTimeout: 30
   receiveTimeout: 60
-  localSite: source-id
\ No newline at end of file
+  localSite: destination-id
```
kcover: Sites need to be regional for the Ion Architecture to work with them, and we only harvest from regional sites so the local site will be the destination.

query/src/main/java/com/connexta/replication/query/QueryService.java
```diff
@@ -170,12 +175,16 @@ private void doQuery() {
...
     for (Filter filter : filters) {
       final FilterIndex filterIndex = queryServiceTools.getOrCreateFilterIndex(filter);
+      LOGGER.debug(
+          "Obtained FilterIndex {}, of filter {}, of site {}", filterIndex, filter, site.getName());
 
       // holds the latest modified time received so far
       AtomicReference<Instant> latestModifiedTime =
-          new AtomicReference<>(filterIndex.getModifiedSince().orElse(Instant.MIN));
+          new AtomicReference<>(
+              filterIndex.getModifiedSince().orElse(new Date(Long.MIN_VALUE).toInstant()));
```
kcover: Solr can only save dates so using Instant.MIN causes an error. I started using the min date instead.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerTest.java
```diff
@@ -251,15 +249,14 @@ void testDestroy() {
...
 
-    // when
-    new WorkerManager(siteManager, properties, threadPoolFactory);
-
-    // then
-    exception.expect(NotFoundException.class);
+    // expect
+    Assertions.assertThrows(
+        NotFoundException.class,
+        () -> new WorkerManager(siteManager, properties, threadPoolFactory));
```
kcover: This test was not catching the exception for me so I used the junit 5 documentation to find this way of expecting an exception.

kcover:
Seeing some wonkiness in the jenkins CI that doesn't make sense to me. build now.

replication-api-impl/src/test/java/com/connexta/replication/api/impl/worker/WorkerManagerTest.java
```diff
@@ -251,15 +249,14 @@ void testDestroy() {
 
   @Test
   void testNoLocalSiteConfigured() {
-    // setup
+    // given
     when(properties.getLocalSite()).thenReturn("");
-    when(siteManager.get(LOCAL_SITE)).thenThrow(NotFoundException.class);
+    when(siteManager.get("")).thenThrow(NotFoundException.class);
```
peterhuffer: Whoops, I think we can just delete the line right above this then put it back to the way it was and it should work.

peterhuffer:
Wow! Classic!

replication-api-impl/src/main/java/com/connexta/replication/api/impl/spring/ServiceConfig.java
```diff
@@ -85,6 +91,26 @@ public NodeAdapters nodeAdapters(List<NodeAdapterFactory> nodeAdapterFactories)
     return nodeAdapters;
   }
 
+  @Profile("Ion")
+  @Bean
+  public QueueBroker queueBroker(MeterRegistry meterRegistry) {
+    return new MemoryQueueBroker(20, meterRegistry);
```
paouelle: ✏️ I think we can use a much higher capacity than this. I would bump this up into the 1000s

query/src/main/java/com/connexta/replication/query/QueryService.java
paouelle: Might be a good test to add to the Solr/spring-data integration test cases that deals with `Instant`

distributions/test/src/test/java/com/connexta/replication/distributions/test/TestUtils.java
kcover: I want to but test containers doesn't provide a way to do that. I could look into a solution but that will be quite a bit of work and I don't want to put anymore effort into it unless it becomes a problem.

PR #261, opened 2019-10-01
REPL-150 Task/resource/metadata impl and pojo implementation

paouelle:
#### What does this PR do?
This PR introduces implementation for the task/resource/metadata along with counterpart pojos which can be used to persist tasks when we moved to an external queues. To do so, it annotates these classes and existing ones with Jackson annotations which we can leverage to generate a Json representation of these objects which should be suitable to send as bytes via RabbitMQ.

It abstracts out task logic from MemoryTask and removes all default constructors for Pojos to ensure Json/Solr deserialization starts fresh as opposed to starting with the current version as it did before. This will allow proper detection of missing attributes.

@kcover 
@peterhuffer 
@clockard 

Successful build which includes E2E, component, and integration test cases

- The resource and metadata info classes were moved from where they were as used by the Query Service to the data package and tweaked to be in line with the other data impl classes we have.
- New corresponding Pojos were added that conforms to those we already created for sites and filters.
- The AbstractTaskImpl is an abstraction of the MemoryTask class so it becomes reusable for the RabiitMQ implementation later.
- Some of the pojos were annotated with Jackson annotations to support persistence to/from Json as will be used with RabbitMQ

Fixes #XXXX

#### Checklist:
- [ ] Documentation Updated
- [X] Update / Add Unit Tests
- [X] Update / Add Integration Tests


replication-api/src/main/java/com/connexta/replication/api/data/TaskInfo.java
```diff
@@ -32,7 +32,7 @@
    *
    * @return the intel document identifier
    */
-  public String getId();
+  public String getIntelId();
```
paouelle: 📔 This was to avoid confusion with the id of an object representing the uniqueness of that object and not the intel identifier.

kcover: ✏️ MetadataId would be more consistent with what we have.

paouelle: This is the identifier for the whole intel which includes the product and the various metadata types. The fact that we correlate that with the metacard id when harvesting (or replicating from DDF) is different.

pom.xml
```diff
@@ -562,6 +598,10 @@
             <groupId>com.github.npathai</groupId>
             <artifactId>hamcrest-optional</artifactId>
         </dependency>
+        <dependency>
```
clockard: Do we actually use this dependency anywhere other than tests?

paouelle: Correct, it is only for tests and the management section is defining the default scope for it to be `test`

pom.xml
```diff
@@ -390,6 +390,36 @@
...
+                <artifactId>jackson-datatype-jdk8</artifactId>
+                <version>${jackson.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.datatype</groupId>
+                <artifactId>jackson-datatype-jsr310</artifactId>
+                <version>${jackson.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.module</groupId>
```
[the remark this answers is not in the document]
paouelle: That seems to have occurred after I rebased. Thanks for noticing.

replication-api-impl/src/main/java/com/connexta/replication/api/impl/data/MetadataInfoImpl.java
```diff
@@ -0,0 +1,238 @@
...
+    }
+    super.writeTo(pojo);
+    setOrFailIfNullOrEmpty("type", this::getType, pojo::setType);
+    setOrFailIfNull("last modified", this::getLastModified, pojo::setLastModified);
+    return pojo.setVersion(MetadataInfoPojo.CURRENT_VERSION).setSize(Math.max(-1L, size));
+  }
+
+  @Override
+  protected final void readFrom(MetadataInfoPojo<?> pojo) {
+    super.readFrom(pojo);
```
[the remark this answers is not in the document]
paouelle: It is up to the class itself to know what they can and cannot deserialize. You would just be pushing that logic elsewhere and introduce something else in the mix to make that decision.

PR #265, opened 2019-10-07
REP-264 Add modified time to Ion ingest requests

### LINKED ISSUE #264: Add metadata modified time to ion ingest requests
**Describe the task**
Add metadata modified time to ion ingest requests.

kcover:
#### What does this PR do?
Adds the metadata modified time to ingest requests for ion.

start the replication stack and ingest something into ion.

Fixes #264 

#### Checklist:
- [x] Documentation Updated
- [x] Update / Add Unit Tests

### DIFF
--- a/adapters/ion-adapter/src/main/java/com/connexta/replication/adapters/ion/IonNodeAdapter.java
+++ b/adapters/ion-adapter/src/main/java/com/connexta/replication/adapters/ion/IonNodeAdapter.java
@@ -124,6 +124,8 @@ public class IonNodeAdapter implements NodeAdapter {
     HttpHeaders headers = new HttpHeaders();
     headers.set("Accept-Version", "0.1.0");
     headers.setContentType(MediaType.MULTIPART_FORM_DATA);
+    headers.set(
+        "Last-Modified", resource.getMetadata().getMetadataModified().toInstant().toString());
 
     HttpEntity<MultiValueMap<String, Object>> requestEntity =
         new HttpEntity(multipartBody, headers);
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,8 +27,7 @@ jobs:
             - mvn-repo-v1-{{ .Branch }}-sonar
             - mvn-repo-v1-{{ .Branch }}-owasp
             - mvn-repo-v1-{{ .Branch }}
-      - setup_remote_docker:
-          docker_layer_caching: true
+      - setup_remote_docker
       - run:
           name: Building
           command: |
@@ -66,8 +65,7 @@ jobs:
       - run: *env
       - checkout
       - restore_cache: *restore_cache
-      - setup_remote_docker:
-          docker_layer_caching: true
+      - setup_remote_docker
       - run:
           name: Building
           command: |
--- a/README.md
+++ b/README.md
@@ -150,26 +150,25 @@ curl -H "Content-Type: application/json" \
 -d @sites.json \
 http://localhost:8983/solr/replication_site/update?commitWithin=1000
 ```
-###### Adding Replication Jobs Example
-Example jobs.json
+###### Adding Replication Filters Example
+Example filters.json
 ```json
-  [
-    {
-      "version": 1,
-      "id": "unique-job-id-98765",
-      "name": "pdf-harvest",
-      "bidirectional": false,
-      "source":" some-unique-id-1234",
-      "destination": "another-unique-id-5678",
-      "filter": "\"media.type\" like 'application/pdf'",
-      "suspended": false
-    }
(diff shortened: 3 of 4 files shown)


adapters/ion-adapter/src/main/java/com/connexta/replication/adapters/ion/IonNodeAdapter.java
```diff
@@ -119,6 +119,7 @@ public boolean createResource(CreateStorageRequest createStorageRequest) {
     builder.part("file", getResourceEntity(resource));
     builder.part("metacard", getMetadataEntity(resource.getMetadata()));
     builder.part("correlationId", resource.getId());
+    builder.part("metadata.modified", resource.getMetadata().getMetadataModified().toString());
```
paouelle: ✏️ Let's not call this metadata as it will get problematic in the long run when a piece of intel can have multiple different type of metadata. This is the modified timestamp for the intel from the perspective of Ion.

What is the expected format? I doubt we want to rely on the `Date.toString()` method to fill this.

The product and all the metadata types could have their own modified timestamp. 

❓ Should Ion store one for each? As such, we would send one for the intel and one for the metacard passed in this message.

We need to figure out which one we associate with the intel from the DDF's perspective such that it is enough to distinguish different versions.

emmberk: ❗️The name of this part needs to exactly match the name and format specified in the IngestAPI, and there isn't a PR to add that new PR yet. This PR should not be merged until that new version of the IngestAPI is released.

paouelle: We understand, this will be done in CDR as part of this first Sprint as agreed during planning right so we can complete this story?

kcover: Isn't each metadata type just another form on the same metadata? In that case each one should have the same metadata modified time.

I think Date.toString() uses ISO format and can easily be parsed by other java applications so I'm not sure if there's anything wrong with it. This should just be used by cdr I think and does need to be to some I2O specification.

paouelle: metadata can be updated/created at different time so we might want to keep track of each individually in the future. For this PR, we care about the intel and should reference the modified timestamp for the intel along with its correlation id already provided.

As for the time format, it is up to the ingest API to dictate what they want to support. That being said, the OpenAPI spec indicates the format should follow RFC3339 (https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#data-types) which is ISO 8601.

kcover:
Reminder to myself to use the date time format specified here https://swagger.io/docs/specification/data-models/data-types/#string

adapters/ion-adapter/src/main/java/com/connexta/replication/adapters/ion/IonNodeAdapter.java
jhunzik: Here's our API change: https://github.com/connexta/ion-ingest-api/commit/b80d72e242f1b5c9540c1ddef7a14ad43d922cc5

kcover:
The modified time format will now look like "2019-10-09T23:50:50.650Z"

adapters/ion-adapter/src/main/java/com/connexta/replication/adapters/ion/IonNodeAdapter.java
```diff
@@ -124,6 +124,8 @@ public boolean createResource(CreateStorageRequest createStorageRequest) {
     HttpHeaders headers = new HttpHeaders();
     headers.set("Accept-Version", "0.1.0");
     headers.setContentType(MediaType.MULTIPART_FORM_DATA);
+    headers.set(
+        "Last-Modified", resource.getMetadata().getMetadataModified().toInstant().toString());
```
mcalcote: ❓ I know this gives us an ISO-8601 representation, but does that mean we get an ISO offsetdatetime string? We expect DateTimeFormat.ISO.DATE_TIME from org.springframework.format.annotation.

kcover: Apparently instant is basically an offsetdatetime with no offset. https://stackoverflow.com/questions/32437550/whats-the-difference-between-instant-and-localdatetime. If I use OffsetDateTime.parse().toString() it works and prints out the same.

PR #268, opened 2019-10-14
Upgrade ReadMe

kcover:
#### What does this PR do?
Adds details to the readme to make it a decent guide for anyone trying to use the replication service.

#### Checklist:
- [x] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,10 +21,12 @@ Derived resources, from products such as NITFs, will not be replicated.
 ## Docker Compose Deployment
 
 #### Prerequisites
-The replication docker stack requires the following to be configured on the swarm before deploying:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,7 +140,28 @@ curl -H "Content-Type: application/json" \
 -d @/path/to/json/config/file.json \
 http://localhost:8983/solr/<target-core>/update?commitWithin=1000
 ``` 
+
+#### Types and Kinds of sites
+Sites can be of different types and kinds. The type and kind of the remote site (the one that's not 
+the local site) will determine whether the replication is a push, pull, harvest or both.
```
tyler30clemens: Instead of just "both" it could be "both push and pull" to avoid confusion because of the third term "harvest".

PR #311, opened 2020-05-14
webHDFS Adapter

cjlange:
#### What does this PR do?
- Adds an adapter for webHDFS

@aaronilovici 
@mdang8
@clockard 

[DIB-9549](https://jira.di2e.net/browse/DIB-9549)

#### Checklist:
- [ ] Documentation Updated
- [x] Update / Add Unit Tests
- [ ] Update / Add Integration Tests --> see #310

### DIFF
--- a/adapters/webhdfs-adapter/pom.xml
+++ b/adapters/webhdfs-adapter/pom.xml
@@ -63,6 +63,12 @@
             <version>2.10.3</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>25.0-jre</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
 
 
--- a/adapters/webhdfs-adapter/pom.xml
+++ b/adapters/webhdfs-adapter/pom.xml
@@ -51,6 +51,12 @@
             <version>${apache.httpcomponents.version}</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.10.3</version>
+            <scope>compile</scope>
+        </dependency>
     </dependencies>
 
 
--- a/adapters/webhdfs-adapter/pom.xml
+++ b/adapters/webhdfs-adapter/pom.xml
@@ -71,5 +71,46 @@
         </dependency>
     </dependencies>
 
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>default-check</id>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
+                        <configuration>
+                            <haltOnFailure>true</haltOnFailure>
+                            <rules>
+                                <rule>
+                                    <element>BUNDLE</element>
+                                    <limits>
+                                        <limit implementation="org.codice.jacoco.LenientLimit">
+                                            <counter>INSTRUCTION</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.94</minimum>
+                                        </limit>
+                                        <limit implementation="org.codice.jacoco.LenientLimit">
+                                            <counter>BRANCH</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.84</minimum>
+                                        </limit>
+                                        <limit implementation="org.codice.jacoco.LenientLimit">
+                                            <counter>COMPLEXITY</counter>
+                                            <value>COVEREDRATIO</value>
+                                            <minimum>0.84</minimum>
+                                        </limit>
+                                    </limits>
+                                </rule>
+                            </rules>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
 
 </project>
--- a/adapters/webhdfs-adapter/pom.xml
+++ b/adapters/webhdfs-adapter/pom.xml
@@ -22,12 +22,14 @@
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
+    <name>Replication :: Adapters :: WebHDFS</name>
+    <groupId>replication-adapters</groupId>
     <artifactId>webhdfs-adapter</artifactId>
+    <packaging>bundle</packaging>
 
     <properties>
         <apache.httpcomponents.version>4.5.2</apache.httpcomponents.version>
         <apache.httpclient.version>4.5.12</apache.httpclient.version>
-        <tika.mime.resolver.version>2.21.4</tika.mime.resolver.version>
     </properties>
 
     <dependencies>
@@ -61,7 +63,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.10.3</version>
+            <version>${jackson.version}</version>
             <scope>compile</scope>
         </dependency>
         <dependency>
@@ -73,7 +75,7 @@
         <dependency>
             <groupId>ddf.mime.tika</groupId>
             <artifactId>mime-tika-resolver</artifactId>
-            <version>${tika.mime.resolver.version}</version>
+            <version>${ddf.version}</version>
         </dependency>
     </dependencies>
 
@@ -115,6 +117,33 @@
                         </configuration>
                     </execution>
                 </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.felix</groupId>
+                <artifactId>maven-bundle-plugin</artifactId>
+                <extensions>true</extensions>
+                <configuration>
+                    <instructions>
+                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <Import-Package>
+                            !org.jvnet.staxex,
+                            ddf.security.permission,
+                            ddf.security.encryption;version="[1.0.0, 3.0.0]",
+                            *
+                        </Import-Package>
+                        <Embed-Dependency>
+                            commons-collections4,
+                            ogc-tools-gml-jts,
+                            ddf-security-common,
+                            mime-tika-resolver,
+                            platform-util,
+                            jaxb-core,
+                            commons-lang3,
+                            commons
+                        </Embed-Dependency>
+                        <Embed-Transitive>true</Embed-Transitive>
+                    </instructions>
+                </configuration>
             </plugin>
         </plugins>
     </build>
--- a/adapters/webhdfs-adapter/pom.xml
+++ b/adapters/webhdfs-adapter/pom.xml
@@ -13,21 +13,16 @@
  **/
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <parent>
-    <artifactId>replication</artifactId>
-    <groupId>replication</groupId>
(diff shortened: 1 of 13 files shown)


cjlange:
build now

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
```diff
@@ -0,0 +1,267 @@
...
+    LOGGER.debug("The location being written to is: {}", location);
+    HttpPut httpPut = new HttpPut(location);
+
+    // only a single resource is supported at this time and is the reason for always retrieving from
+    // index zero
+    Resource resource = createStorageRequest.getResources().get(0);
+
+    MultipartEntityBuilder builder = MultipartEntityBuilder.create();
+    builder.addBinaryBody(
+        "file", resource.getInputStream(), ContentType.create(resource.getMimeType()), "file.ext");
```
clockard: Should this be resource.getName() instead of "file.ext"

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
```diff
@@ -0,0 +1,267 @@
...
+                String.format("Request failed with status code: %d", status));
+          }
+        };
+
+    return sendHttpRequest(httpPut, responseHandler);
+  }
+
+  @Override
+  public boolean updateResource(UpdateStorageRequest updateStorageRequest) {
+    return false;
```
clockard, line 295: I know there are more iterations of this coming but I was wondering if there is any reason not to implement this by simply calling the createResource method.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
```diff
@@ -0,0 +1,267 @@
...
+  /**
+   * Sends a PUT request with no file data and without following redirects, in order to retrieve the
+   * location to write to
+   *
+   * @param createStorageRequest request containing the {@link
+   *     org.codice.ditto.replication.api.data.Resource} to create
+   * @return a {@code String} containing the location to write to
+   * @throws URISyntaxException when the fileUrl cannot be parsed as a URI reference
+   */
+  @VisibleForTesting
```
mdang8, line 171: ❓ Do you need this annotation since the method is already package private?

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
```diff
@@ -0,0 +1,267 @@
...
+    this.client = client;
+  }
+
+  @Override
+  public boolean isAvailable() {
+    LOGGER.debug("Checking access to: {}", webHdfsUrl);
+
+    try {
+      URIBuilder builder = new URIBuilder(webHdfsUrl.toString());
+      builder.setParameter("op", "CHECKACCESS").setParameter("fsaction", "ALL");
```
mdang8: ❗ Turns out the `fsaction` parameter doesn't work, at least for our configuration. I was able to verify in testing that having it results in a 400 status code, while removing it gives the correct 200.
```suggestion
      builder.setParameter("op", "CHECKACCESS");
```

cjlange: I removed the parameter, as suggested.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
cjlange: The only reason this is package private is so that I can call the method directly from the unit tests.  While the annotation is purely informational, it is useful for anyone unfamiliar with the code looking at access levels.

cjlange:
build now

clockard (approved):

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
cjlange: Yes.  Updated it.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
cjlange: We certainly could have it do that now, but the implementation work for createResource is happening this week and will not be identical to calling the `createResource` method, so I'd prefer to just leave it like this for this pull request.

aaronilovici: At first glance, the updateResource was going to do more than call createResource. It was first going to do a GET on the filesystem and see if the file we're trying to update already exists.

But, since updateResource will only be called if the resource is indeed updated, then we will have a new resource modified timestamp and the file name will indeed be different by the suffix. IE, fileA_ts1 exists, we get update with ts2, then we create fileA_ts2 on the filesystem.

As long as we have overwrite=false when we're calling createResource from updateResource, we don't need to do any other implementation to get this to work.

Lange - I agree, though. We have a ticket for this and we should use it, because it will also be good for an integration test to make sure we aren't overwriting files. So for the combined effort I think we should keep it in the second ticket for this sprint.

adapters/webhdfs-adapter/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+    <properties>
+        <apache.httpcomponents.version>4.5.2</apache.httpcomponents.version>
+        <apache.httpclient.version>4.5.12</apache.httpclient.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>replication</groupId>
+            <artifactId>replication-api</artifactId>
+            <version>0.3.0-SNAPSHOT</version>
```
aaronilovici: This should be `${project.version}` to make sure that it's always using the latest version we're building.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
```diff
@@ -0,0 +1,267 @@
...
+  public boolean createResource(CreateStorageRequest createStorageRequest) {
+    try {
+      String location = getLocation(createStorageRequest);
+
+      return writeFileToLocation(createStorageRequest, location);
+
+    } catch (URISyntaxException e) {
+      LOGGER.error("Unable to get location, due to invalid URL.", e);
+    } catch (ReplicationException e) {
+      LOGGER.error("Unable to create resource.", e);
```
aaronilovici, line 157: Could be helpful here to add the name to the string so we know what we'd be looking for in a dib installation.

LOGGER.error(String.format("Unable to create resource %s), createStorageRequest.getResources().get(0).getName()), e)

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
```diff
@@ -0,0 +1,267 @@
...
+
+  @Override
+  public boolean createResource(CreateStorageRequest createStorageRequest) {
+    try {
+      String location = getLocation(createStorageRequest);
+
+      return writeFileToLocation(createStorageRequest, location);
+
+    } catch (URISyntaxException e) {
+      LOGGER.error("Unable to get location, due to invalid URL.", e);
```
aaronilovici, line 155: We could included the invalid URL in the error message to make it easier to debug later if needed.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
```diff
@@ -0,0 +1,267 @@
...
+   *
+   * @param createStorageRequest request containing the {@link
+   *     org.codice.ditto.replication.api.data.Resource} to create
+   * @return a {@code String} containing the location to write to
+   * @throws URISyntaxException when the fileUrl cannot be parsed as a URI reference
+   */
+  @VisibleForTesting
+  String getLocation(CreateStorageRequest createStorageRequest) throws URISyntaxException {
+
+    if (createStorageRequest.getResources().get(0) == null) {
```
aaronilovici: Could also do if(createStorageRequest.getResources().isEmpty()) { }

Not sure if calling get(0) would throw an NPE if there is nothing in that index. But that's something Matt's integration test would show, so probably not a huge deal if this works as expected with an empty list.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
cjlange: Upon further thought/discussion, this is changing slightly so that the filename used here matches the filename used in the URL.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
cjlange: As followup, there will be a lot of overlap, but I'm going to leave that work for the separate ticket we have created, so as to avoid prolonging this pull request.

adapters/webhdfs-adapter/pom.xml
cjlange: Updated.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
cjlange: Added the name into the error logging.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
cjlange: Added an empty list check as well.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
cjlange: Leaving this as is for now, as discussed.

adapters/webhdfs-adapter/src/main/java/com/connexta/replication/adapters/webhdfs/WebHdfsNodeAdapter.java
cjlange: Leaving this as is for now, as discussed.

cjlange:
build now

cjlange:
build now

cjlange:
build now

cjlange:
build now

mdang8 (approved):

aaronilovici:
Successful hero! After configuring the cluster correctly (webhdfs enabled, permissions disabled per CDF), I was able to set up replication with this branch installed in my DIB instance, and watch files get replicated (one way only, as query is not implemented in this adapter to be able to support bidirectional replications) to my local HDFS instance. The logs were free of errors.

aaronilovici (approved):

PR #348, opened 2020-10-14
Update replication to work with DDF 2.23.0

clockard:
#### What does this PR do?
Updates replication so it will work with DDF 2.23.0 and newer systems
Second commit is to cache the clients as it was taking several seconds (sometimes 10+) to create the client.

@kcover 
@cjlange 

Install and verify basic replication works. 
I was using an alliance 1.10.1 for testing which uses DDF 2.23.0

I have manually tested this on alliance 1.10.1 and ddf 2.25.2-SNAPSHOT


clockard:
build now

clockard:
build now

clockard:
build now

adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
```diff
@@ -169,9 +169,11 @@ public String getSystemName() {
...
     }
 
-    String tagFilter = CqlBuilder.equalTo(Constants.METACARD_TAGS, Constants.REGISTRY_TAG);
-    String identityFilter = CqlBuilder.negate(CqlBuilder.isNull(Constants.REGISTRY_IDENTITY_NODE));
-    String filter = CqlBuilder.allOf(tagFilter, identityFilter);
+    // get any one metacard from the system
+    String modifiedFilter = CqlBuilder.negate(CqlBuilder.isNull(Constants.METACARD_MODIFIED));
+    String tagFilter = CqlBuilder.like(Constants.METACARD_TAGS, "*");
+    String noTagFilter = CqlBuilder.negate(CqlBuilder.isNull(Constants.METACARD_TAGS));
+    String filter = CqlBuilder.anyOf(modifiedFilter, tagFilter, noTagFilter);
```
kcover, line 176: ❓ Wouldn't the modified filter be enough to get any metacard?

admin/query/pom.xml
```diff
@@ -79,6 +79,12 @@
                         <Embed-Dependency>
                             admin-query-common
                         </Embed-Dependency>
+                        <Import-Package>
+                          org.codice.ddf.admin.configurator;version="[2.23,3)",
+                          org.codice.ddf.configuration;version="[2.23,3)",
+                          org.codice.ddf.internal.admin.configurator.actions;version="[2.23,3)",
+                            *
```
kcover, line 86: ❓ Why the need to import ddf artifacts where we didn't before?

adapters/pom.xml
```diff
@@ -26,6 +26,6 @@
 
   <modules>
     <module>ddf-adapter</module>
-    <module>webhdfs-adapter</module>
+<!--    <module>webhdfs-adapter</module>-->
```
kcover, line 29: ❓ Why did we have to remove this?

kcover (approved):

adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/DdfNodeAdapter.java
clockard: If the system is "new" and doesn't have any resource data in it but does have other metacards like registry, geoname, etc then we can use those as well to get the site id which is why the metacard tags part.

admin/query/pom.xml
clockard: This is what is letting us use 2.23.0 ddf dependencies instead of requiring the 2.25.2 ones we build with. If we don't specify the imports it defaults to [2.25.2,3). This gives us the backwards compatibility needed to work with alliance 1.10.x

adapters/pom.xml
clockard: We had removed it from the feature files when upgrading to DDF 2.25.x. It will take additional work to get the webhdfs adapter working with 2.25.x and address owasp findings.

clockard:
build now

PR #352, opened 2020-11-12
Run updates

clockard:
#### What does this PR do?
Adds the ability to run replication jobs from the ui
Adds the ability to again run replication jobs from the command line.
Re-enables the webhdfs adapter

@kcover @peterhuffer if/when you guys have time. I was doing this on my own time because I think it is useful

Install and verify that jobs can be run from the command line and from the UI

### DIFF
--- a/adapters/webhdfs-adapter/pom.xml
+++ b/adapters/webhdfs-adapter/pom.xml
@@ -16,7 +16,7 @@
     <parent>
         <artifactId>adapters</artifactId>
         <groupId>replication</groupId>
-        <version>0.3.2-SNAPSHOT</version>
+        <version>0.3.3-SNAPSHOT</version>
     </parent>
     <modelVersion>4.0.0</modelVersion>
 
--- a/adapters/webhdfs-adapter/pom.xml
+++ b/adapters/webhdfs-adapter/pom.xml
@@ -29,7 +29,7 @@
         <apache.httpcomponents.version>4.5.2</apache.httpcomponents.version>
         <apache.httpclient.version>4.5.12</apache.httpclient.version>
         <guava.version>20.0</guava.version>
-        <hadoop.version>2.7.3</hadoop.version>
+        <hadoop.version>2.10.1</hadoop.version>
         <hadoop.mini.clusters.version>0.1.16</hadoop.mini.clusters.version>
     </properties>
 
@@ -87,6 +87,7 @@
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-minicluster</artifactId>
             <version>${hadoop.version}</version>
+            <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>replication</groupId>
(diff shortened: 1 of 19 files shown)


clockard:
build now

clockard:
build now

clockard:
build now

clockard:
build now

clockard:
build now

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
```diff
@@ -33,7 +33,9 @@
 
 /**
  * The ReplicatorRunner periodical queues up replication jobs for all the current replication
- * configurations.
+ * configurations. If the configured replication period is less than or equal to zeor the
```
aaronilovici: nitpick: zeor -> zero

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/ReplicatorRunnerTest.java
```diff
@@ -91,6 +91,15 @@ public void initNonDefaultPeriod() {
     assertThat(period.getValue(), is(1L));
   }
 
+  @Test
+  public void initZeorPeriod() {
```
aaronilovici: initZeorPeriod -> initZeroPeriod

adapters/webhdfs-adapter/pom.xml
```diff
@@ -29,7 +29,7 @@
         <apache.httpcomponents.version>4.5.2</apache.httpcomponents.version>
         <apache.httpclient.version>4.5.12</apache.httpclient.version>
         <guava.version>20.0</guava.version>
-        <hadoop.version>2.7.3</hadoop.version>
+        <hadoop.version>2.10.1</hadoop.version>
```
aaronilovici, line 32: I'm not sure this will run with hadoop 2.10.x, we can only test it with a local cluster of the same version. In the itests, the mini hdfs cluster is used which has a different version of hadoop and some of the webHDFS calls are structured differently. This is fine because we won't be using the HDFS adapter, but I'm not sure it'll work as expected when switching versions if someone actually wanted to use it.

aaronilovici (approved):
LGTM, small points of feedback.

clockard:
build now

replication-api-impl/src/main/java/org/codice/ditto/replication/api/impl/ReplicatorRunner.java
clockard: Fixed

replication-api-impl/src/test/java/org/codice/ditto/replication/api/impl/ReplicatorRunnerTest.java
clockard: fixed

adapters/webhdfs-adapter/pom.xml
clockard: I wasn't sure about the webhdfs stuff other than getting it to build and pass it's tests. I will remove it from the build again for now.

clockard:
build now

PR #418, opened 2021-10-18
NA-47 Location attribute doesn't get replicated over for metacards without resource

sgalla:
#### What does this PR do?
Fixed two issues: 
1. The location attribute wasn't always getting replicated over when replicating a metacard with no resource. This was due to logic expecting the `geometry` xml tag to not be namespaced. Because it can be namespaced (for example, <ns2:geometry> instead of just <geometry>), I added logic to trim off the namespace before doing the existing check.

2. Added a webClient.reset() call in the get method of DdfRestClient. Because we are using the same WebClient instance, without a call to reset, the path and query parameters kept appending to one another on subsequent calls to the get method, creating invalid URLs. 

@clockard 
@shaundmorris 
@rececoffin

1. Replicate a few metacards that have location attributes and no resources attached and verify the location values are properly replicated. 
2. Replicate a few metacards that do have resources attached and verify that they all get replicated over.

https://octoconsulting.atlassian.net/browse/NA-47

#### Checklist:
- [ ] Documentation Updated
- [ ] Update / Add Unit Tests

### DIFF
--- a/adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/csw/Constants.java
+++ b/adapters/ddf-adapter/src/main/java/com/connexta/replication/adapters/ddf/csw/Constants.java
@@ -13,6 +13,9 @@
  */
 package com.connexta.replication.adapters.ddf.csw;
 
+import java.util.Arrays;
+import java.util.List;
+
 public final class Constants {
 
   private Constants() {}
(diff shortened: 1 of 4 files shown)


clockard (approved):
Thanks for fixing these!

clockard:
Verified webClient.reset() ✅

